PR #3, opened 2016-08-08
Update build.rs

keorn:
fixes serde errors

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -12,10 +12,7 @@ mod inner {
         let src = Path::new("src/spec/mod.rs.in");
         let dst = Path::new(&out_dir).join("mod.rs");
 
-        let mut registry = syntex::Registry::new();
-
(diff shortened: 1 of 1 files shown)


debris:
travis failing

keorn:
Yep, I saw that. Trying to figure it out, but not sure about the serde workings. When I included ethabi as dependency without this commit it was failing to compile, but now it fails standalone. It comes from the zkSNARKs guys: https://github.com/gancherj/ethabi

debris:
It's caused by locked dependencies in `Cargo.lock`. Unfortunately updating dependencies is not enough, cause some tests are failing with new version of serde. I will fix them and merge the pr

keorn:
Ah I see, thanks.

debris:
related issue: https://github.com/serde-rs/serde/issues/487

debris:
@keorn please review #4

PR #4, opened 2016-08-09
dependencies update

debris:
updated deps and fixed tests which failed due to regression in serde codegen

### DIFF
--- a/src/spec/operation.rs
+++ b/src/spec/operation.rs
@@ -19,20 +19,17 @@ pub enum Operation {
 impl Deserialize for Operation {
 	fn deserialize<D>(deserializer: &mut D) -> Result<Self, D::Error>
 		where D: Deserializer {
-		let v = try!(Value::deserialize(deserializer));
-		if let Value::Object(ref map) = v.clone() {
-			if let Some(&Value::String(ref s) = map.get("type") {
-				let result = match s.as_ref() {
-					"constructor" => Deserialize::deserialize(&mut value::Deserializer::new(v)).map(Operation::Constructor),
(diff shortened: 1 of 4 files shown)


keorn:
Seems to work, but now I get the same error with `ethstore`, I guess it needs a similar pr:
```
   Compiling ethstore v0.1.0 ()                                   │
ethstore/build.rs:33:33: 33:46 error: mismatched types [E0308]                                       │
ethstore/build.rs:33         serde_codegen::register(&mut registry);                                 │
                                                     ^~~~~~~~~~~~~                                   │
ethstore/build.rs:33:33: 33:46 help: run `rustc --explain E0308` to see a detailed explanation       │
ethstore/build.rs:33:33: 33:46 note: expected type `&mut syntex::Registry`                           │
ethstore/build.rs:33:33: 33:46 note:    found type `&mut inner::syntex::Registry`                    │
error: aborting due to previous error  
```

debris:
I'm on it

src/spec/operation.rs
```diff
@@ -19,20 +19,17 @@ pub enum Operation {
 impl Deserialize for Operation {
 	fn deserialize<D>(deserializer: &mut D) -> Result<Self, D::Error>
 		where D: Deserializer {
-		let v = try!(Value::deserialize(deserializer));
-		if let Value::Object(ref map) = v.clone() {
-			if let Some(&Value::String(ref s) = map.get("type") {
```
oli-obk: how did that ever work? you were missing a bracket

debris: you are right! I have no idea! :)

PR #26, opened 2017-05-24
Will you change EventInterface to public?

usd-yamazaki:
I want to execute this code.

    let abi = ethabi::Contract::new(ethabi::Interface::load(json.as_slice()).unwrap());
    let events = abi.events().map(|event| {
        let params = event.interface.inputs.iter().map(|event_param| {
            event_param.kind.clone()
        } ).collect::<Vec<_>>();
        (H256::from(event_signature(&event.interface.name, params.as_slice()).as_ref()), event)
    });

Thank you.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -1,7 +1,7 @@
 //! Contract event.
 
 use std::collections::HashMap;
-use spec::{Event as EventInterface, ParamType};
+use spec::{Event as EventInterface, ParamType, EventParam};
 use decoder::Decoder;
 use token::Token;
 use error::Error;
@@ -100,9 +100,9 @@ impl Event {
 		&self.interface.name
 	}
 
-	/// EventInterface getter
-	pub fn interface(&self) -> &EventInterface {
-		&self.interface
+	/// Return the inputs of the event.
+	pub fn inputs(&self) -> &Vec<EventParam> {
+		&self.interface.inputs
 	}
 }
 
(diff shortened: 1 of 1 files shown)


rphmeier:
Would prefer to expose the fields of `EventInterface` through public getters on `Event` which gives us a lot more room for changes without breaking backwards-compatibility.

usd-yamazaki:
Is this correct?

rphmeier:
@usd-yamazaki Not exactly -- what I mean is that we should leave the `Event` type definition general enough to be implemented without `EventInterface` existing (since it could be removed in the future) while maintaining backwards compatibility.

For example, if you want to call `event.interface.inputs.iter()`, you could write a `fn inputs(&self) -> &[EventParam]` which does `&self.interface.inputs` internally.

Then your code becomes `event.inputs().iter()` which will behave identically.
There is already a getter for `name`.

usd-yamazaki:
Thanks for the explanation. I think I understood it.

src/event.rs
```diff
@@ -133,6 +134,11 @@ impl Event {
 	pub fn name(&self) -> &str {
 		&self.interface.name
 	}
+
+	/// Return the inputs of the event.
+	pub fn inputs(&self) -> &Vec<EventParam> {
```
rphmeier: You can change `&Vec<EventParam>` to `&[EventParam]`. It's more idiomatic

rphmeier:
Awesome, just one tiny nitpick and then I'm happy to merge.

usd-yamazaki:
Thank you for telling me.
I changed it to shortened form.

PR #27, opened 2017-06-13
Index contract operations

### LINKED ISSUE #20: Don't linear search for operations on each lookup
It would probably be faster to put ops in a hash map (maybe even a PHF map) by name and look them up in O(1) time.

andrebeat:
Index the contract operations by name (and the constructor separately) to avoid performing linear search on each lookup. Fixes #20.

### DIFF
--- a/src/contract.rs
+++ b/src/contract.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use spec::{Interface, Operation};
 use function::Function;
 use constructor::Constructor;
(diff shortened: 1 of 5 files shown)


andrebeat:
Should I remove the lookup methods from `Interface`? I'd suggest to create iterators for functions and events in `Interface` and use those when creating `Contract`.

src/contract.rs
```diff
@@ -7,49 +9,69 @@ use error::Error;
 /// API building calls to contracts ABI.
 #[derive(Clone, Debug)]
 pub struct Contract {
-	interface: Interface,
+    constructor: Option<ConstructorInterface>,
+    functions: HashMap<String, FunctionInterface>,
+    events: HashMap<String, EventInterface>,
```
debris: maybe change to:

```
constructor: Option<Constructor>,
functions: HashMap<String, Function>,
events: HashMap<String, Event>,
```

debris:
yes, you can remove them ;)

src/contract.rs
andrebeat: Done.

debris:
lgtm, @rphmeier what do you think?

src/spec/interface.rs
```diff
@@ -13,30 +13,6 @@ impl Interface {
 		serde_json::from_slice(bytes).map_err(From::from)
 	}
 
-	/// Returns contract constructor specification.
-	pub fn constructor(&self) -> Option<Constructor> {
```
rphmeier: If you leave these functions the changes are still backwards-compatible. Otherwise, this will need a bump to 2.0

andrebeat: Fixed, I don't think it's worth it to bump a major version now. I added a TODO in the code about removing the functions in a future major version (I can remove it though if you think it's unnecessary).

debris:
thanks!

PR #28, opened 2017-06-13
Ignore signature when parsing function names

### LINKED ISSUE #10: Fix parsing of function names containing the whole signature
More details: https://github.com/ethcore/parity/issues/4122#issue-199917015

andrebeat:
Fixes #10.

### DIFF
--- a/src/spec/operation.rs
+++ b/src/spec/operation.rs
@@ -133,4 +133,31 @@ mod tests {
 		test_sanitize_function_name("()", "");
 		test_sanitize_function_name("", "");
 	}
+
+	#[test]
+	fn deserialize_sanitize_event_name() {
+		fn test_sanitize_event_name(name: &str, expected: &str) {
+			let s = format!(r#"{{
+				"type":"event",
+					"inputs": [{{
+						"name":"a",
+						"type":"address",
+						"indexed":true
+					}}],
+					"name":"{}",
+					"outputs": [],
+					"anonymous": false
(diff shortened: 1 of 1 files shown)


src/spec/operation.rs
```diff
@@ -27,7 +27,14 @@ impl<'a> Deserialize<'a> for Operation {
...
 			"constructor" => from_value(v).map(Operation::Constructor),
-			"function" => from_value(v).map(Operation::Function),
+			"function" => from_value(v).map(|mut f: Function| {
+				// This is a workaround to support non-spec compliant function names,
+				// see: https://github.com/paritytech/parity/issues/4122
+				if let Some(i) = f.name.find('(') {
+					f.name.truncate(i);
+				}
+				Operation::Function(f)
+			}),
```
debris: What about `event` and `fallback`? Shouldn't we also truncate their names?

cc @tomusdrw

tomusdrw: Yeah, IMHO we should do it for `events` as well. Not sure about fallback though, in this functions seems that it's already parsed.

@andrebeat Could you add some more tests for fallback (`name: ""` and `name: "()"`) and events?

andrebeat: The `fallback` event is currently ignored (also according to the spec it doesn't have a name). I just added the same changes for `events` and added more tests. 👍

debris:
thanks! :)

PR #30, opened 2017-06-15
Type check function and constructor calls

### LINKED ISSUE #29: Validate arguments when encoding function calls
Currently the provided function arguments (tokens) are not validated against the function interface. An error should be returned if the types of the provided arguments do not match the interface.

andrebeat:
Fixes #29.


andrebeat:
I've used `Error::InvalidData` for type check errors but maybe it would make sense to have a `TypeCheckError` with more information.

debris:
imo, `InvalidData` is fine :)

PR #37, opened 2017-07-04
Relax required types where possible

Vurich:
A lot of the functions that return or take a `Vec` could be relaxed to take a `&[u8]` or return a `T: FromIterator<u8>`. This commit does that. This is a breaking change, but a minor one. If we don't want to make breaking changes I can write versions of the functions that have the relaxed signature and have the existing functions call into them.

This is part of converting `Vec<u8>` to `Bytes`, which is a prelude to the future queue overhaul.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -173,7 +173,7 @@ mod tests {
 	#[test]
 	fn encode_address() {
 		let address = Token::Address([0x11u8; 20]);
-		let encoded: Vec<_> = encode(vec![address]);
+		let encoded: Vec<_> = encode(&[address]);
 		let expected = "0000000000000000000000001111111111111111111111111111111111111111".from_hex().unwrap();
 		assert_eq!(encoded, expected);
 	}
@@ -183,7 +183,7 @@ mod tests {
 		let address1 = Token::Address([0x11u8; 20]);
 		let address2 = Token::Address([0x22u8; 20]);
 		let addresses = Token::Array(vec![address1, address2]);
-		let encoded: Vec<_> = encode(vec![addresses]);
+		let encoded: Vec<_> = encode(&[addresses]);
 		let expected = ("".to_owned() +
 			"0000000000000000000000000000000000000000000000000000000000000020" +
 			"0000000000000000000000000000000000000000000000000000000000000002" +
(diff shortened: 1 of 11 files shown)


src/encoder.rs
```diff
@@ -110,74 +111,69 @@ impl Mediate {
 	}
 }
 
-/// ABI encoder.
-pub struct Encoder;
+/// Encodes vector of tokens into ABI compliant vector of bytes.
+pub fn encode<T: FromIterator<u8>>(tokens: Vec<Token>) -> T {
```
tomusdrw: Why would you ever want anything else than `Vec<u8>`?
Seems that only `LinkedList` and `VecDeque` from std implementations of  `FromIterator` make any sense, but I don't think we will ever use those for anything (we usually just create a transaction and use this encoded bytes as `data`).

I think this change introduces unecessary confusion and type annotations for little or no gain.

src/function.rs
```diff
@@ -32,21 +33,21 @@ impl Function {
 	}
 
 	/// Prepares ABI function call with given input params.
-	pub fn encode_call(&self, tokens: Vec<Token>) -> Result<Vec<u8>, Error> {
+	pub fn encode_call<T: FromIterator<u8>>(&self, tokens: Vec<Token>) -> Result<T, Error> {
```
tomusdrw: Same here.

src/encoder.rs
rphmeier: It _might_ enable us to collect into a vector created with a custom allocator when those land on stable. But I don't see how that could be done without also passing an allocator reference.

tomusdrw: Just realized that maybe it's because we can get `bytes::Bytes` directly from that function, but it has an efficient `From<Vec<u8>>` implementation anyway, so IMHO sacrificing ease of use of `ethabi` library for that is not worth it.

Vurich: It's efficient, but it's not totally zero-cost. If the encoded value is small the `From` impl wastes space. I also was under the impression that it had to do an alloc when you cloned from a `kind == KIND_VEC` value in order to create the shared buffer but after looking over the code it's clear that that's incorrect. The only difference is when it's inline. It also allows us to use `smallvec` or `elastic-array`. I would frankly have it return an `impl Iterator` so you can do it lazily but obviously that's not stable.

rphmeier: Yep, I think changing to `tokens: &[Token]` or `I: IntoIterator<Item=&Token>` would almost definitely be a bigger performance win than this. This would save us a useless allocation on each `native_contracts` call.

Vurich: I'd be happy to do that. I originally did make the change but the whole stack assumes that it can take an owned `Token` and it wasn't strictly related to the work I was doing so I didn't bother.

rphmeier: sure, but it should definitely be done before releasing 3.0.0

Vurich: Done

rphmeier:
LGTM, but tests will need fixing.

PR #41, opened 2017-08-09
use error-chain

debris:
Experiment to see if error handling with error-chain is easier. It is, by far


Cargo.lock
```diff
@@ -1,8 +1,9 @@
 [root]
 name = "ethabi-cli"
-version = "2.0.0"
+version = "2.0.1"
```
tomusdrw: Isn't it a breaking change? The error types are now different, no?

debris: It is,  I must have modified version before I started working on error-chain changes and forgot to bump it to `3.0`

PR #42, opened 2017-08-16
use_contract!

### LINKED ISSUE #38: Bring `native_contracts` generator to Ethabi repo
...along with a switch to [quote](https://docs.rs/quote) for codegen and breaking dependency on `ethcore-util`. This will ease the transition to procedural-macro based native contract wrappers when macros are more evolved.

The hard part is abstracting over functions that return lists of items: we have some contracts that return a `Vec<Address>` currently, and returning a `Vec<[u8; 20]>` instead would break compatibility and require us to reallocate when converting. `FromIterator<[u8; 20]>` could be used, but there's no implementation of `FromIterator<T> for Vec<U> where U: From<T>` (at least until specialization stabilizes), so we'd have to write wrapper types in Parity.

debris:
fixes #38


ethabi/src/contract.rs
```diff
@@ -0,0 +1,124 @@
...
+			}
+		}
+
+		Ok(result)
+	}
+}
+
+impl Contract {
+	/// Loads contract from json.
+	pub fn load<T: io::Read>(reader: T) -> errors::Result<Self> {
```
rphmeier: indentation of this function

ethabi/src/contract.rs
```diff
@@ -0,0 +1,124 @@
...
+        Functions(self.functions.values())
+    }
+
+    /// Iterate over all events of the contract in arbitrary order.
+    pub fn events(&self) -> Events {
+        Events(self.events.values())
+    }
+
+	/// Returns true if contract has fallback
+	pub fn fallback(&self) -> bool {
```
rphmeier: indentation

derive/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+		})
+		.collect();
+
+	quote! {
+		pub struct #name<'a> {
+			event: &'a ethabi::Event,
+		}
+
+		#[doc(hidden)]
+		impl<'a> From<&'a ethabi::Event> for #name<'a> {
```
rphmeier: these `From` implementations, even though they're hidden, still compromise the type safety of the wrapper. i think having a special `pub(module) fn from_x` for each of these is a better approach

derive/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+					..Default::default()
+				};
+
+				self.event.create_filter(raw).unwrap()
+			}
+		}
+	}
+}
+
+fn declare_functions(function: &Function) -> quote::Tokens {
```
rphmeier: i don't think this does enough -- the function structs declared here only produce the encoded input data but don't seem to handle decoding of output data. if we want to replace the `native_contracts` stuff with this, then we'll need it

ethabi/src/filter.rs
```diff
@@ -0,0 +1,112 @@
+use serde::{Serialize, Serializer};
+use serde_json::Value;
+use hex::ToHex;
+use {Hash, Token};
+
+/// Raw topic filter.
+#[derive(Debug, PartialEq, Default)]
+pub struct RawTopicFilter {
```
rphmeier: this API (especially with the generated event filters) is a little unwieldy to use. IMO a trait-based approach could be cleaner. `struct RawTopicFilter<A=Any, B=Any, C=Any> where A: Topic, B: Topic, C: Topic`. 

where `Topic` is a trait implemented for `struct Any`, `T`, and `Vec<T>`.
Then calling `create_filter`on the generated code would be more like 

`my_contract.events().transfer().create_filter(from, vec![to1, to2])`

ethabi/src/contract.rs
debris: Actually this one is fine. The rest is (and was) broken. I must have used different indentation in the past :/

ethabi/src/filter.rs
debris: > this API (especially with the generated event filters) is a little unwieldy to use. IMO a trait-based approach could be cleaner. `struct RawTopicFilter<A=Any, B=Any, C=Any> where A: Topic, B: Topic, C: Topic`. 

It might better, but then it would be also not consistent with the rest of the crate.

> my_contract.events().transfer().create_filter(from, vec![to1, to2])

It is exactly like that! `RawTopicFilter` is used not exposed in code generated with `use_contract!`.

----

eg. eip20 `Transfer` looks like this:

```json
  {
    "anonymous": false,
    "inputs": [
      {
        "indexed": true,
        "name": "from",
        "type": "address"
      },
      {
        "indexed": true,
        "name": "to",
        "type": "address"
      },
      {
        "indexed": false,
        "name": "value",
        "type": "uint256"
      }
    ],
    "name": "Transfer",
    "type": "event"
  }
```

the generated code is:

```rust
// Struct ethabi_tests::eip20::events::Transfer

fn create_filter(&self, from: Topic<Address>, to: Topic<Address>) -> TopicFilter;

// Struct ethabi_tests::eip20::logs::Transfer

pub struct Transfer {
    pub from: Address,
    pub to: Address,
    pub value: Uint,
}
```

and can be used like you described:

```rust
eip20.events().transfer().create_filter(Topic::Any, Topic::OneOf(vec![to1, to2]));
```

ethabi/src/contract.rs
debris: this one is actually good. The rest was incorrect

derive/src/lib.rs
debris: done in https://github.com/paritytech/ethabi/pull/42/commits/3e366f0e5fb11ca9e65c30f3f64c83af575fa7a1

debris:
also implemented decode for constant functions and contract constructors

ethabi/src/filter.rs
rphmeier: right, I saw that, I just mean that this `Topic::X` API is pretty unwieldy. I think the inner values should be passable by themsevles

derive/src/lib.rs
```diff
@@ -0,0 +1,513 @@
...
+		}
+
+		impl Default for #name {
+			fn default() -> Self {
+				#name::new()
+			}
+		}
+
+		impl #name {
+			pub fn new() -> Self {
```
tomusdrw: What's the point of having `::new()` if you have `::default()`?

derive/src/lib.rs
```diff
@@ -0,0 +1,513 @@
...
+				let contract = ethabi::Contract::load(include_bytes!(#file_path) as &[u8]).unwrap();
```
tomusdrw: That's a runtime failure, the JSON file should be compiled into the binary or the method should return a possible error.

derive/src/lib.rs
```diff
@@ -0,0 +1,513 @@
...
+					ethabi::Token::FixedArray(v)
+				}
+			}
+		},
+	}
+}
+
+fn from_token(kind: &ParamType, token: &syn::Ident) -> quote::Tokens {
+	match *kind {
+		ParamType::Address => quote! { #token.to_address().unwrap() },
```
tomusdrw: why are we sure it will never fail? Can we have a shared PROOF and `expect(PROOF)` here instead?

derive/src/lib.rs
```diff
@@ -0,0 +1,513 @@
...
+		.collect();
+
+	let topic_kinds: Vec<_> = event.inputs
+		.iter()
+		.filter(|param| param.indexed)
+		.map(|param| rust_type(&param.kind))
+		.collect();
+
+	let params: Vec<_> = topic_names.iter().zip(topic_kinds.iter())
+		.map(|(param_name, kind)| quote! { #param_name: ethabi::Topic<#kind> })
```
tomusdrw: Would be cool to have `T: Into<ethabi::Topic<#kind>>` here to allow more flexible API.

derive/src/lib.rs
```diff
@@ -0,0 +1,513 @@
...
+			syn::Ident::new(format!("param{}", index))
+		} else {
+			param.name.to_snake_case().into()
+		}).collect();
+	let kinds: Vec<_> = function.inputs
+		.iter()
+		.map(|param| rust_type(&param.kind))
+		.collect();
+	let params: Vec<_> = names.iter().zip(kinds.iter())
+		.map(|(param_name, kind)| quote! { #param_name: #kind })
```
tomusdrw: Would be awesome to have `T: Into<#kind>` as well.

debris:
> Looks good but, the PR is quite huge. Is it possible to split it into two PRs?

@tomusdrw not really. I've also updated many `ethabi` interfaces and they are used by `ethabi_derive` crate.

ethabi/src/decoder.rs
```diff
@@ -39,190 +34,186 @@ fn as_bool(slice: &[u8; 32]) -> Result<bool, Error> {
...
-					tokens.push(res.token);
-				}
-
-				let result = DecodeResult {
-					token: Token::FixedArray(tokens),
-					new_offset: new_offset,
-				};
-
-				Ok(result)
+fn decode_param(param: &ParamType, slices: &Vec<[u8; 32]>, offset: usize) -> Result<DecodeResult, Error> {
```
debris: yes, refactoring into `Slices`/`Chunks` might be a good idea

ethabi/src/filter.rs
```diff
@@ -0,0 +1,112 @@
...
+impl Serialize for TopicFilter {
+	fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+	where S: Serializer {
+		vec![&self.topic0, &self.topic1, &self.topic2, &self.topic3].serialize(serializer)
+	}
+}
+
+/// Acceptable topic possibilities.
+#[derive(Debug, PartialEq)]
+pub enum Topic<T> {
```
debris: Agreed

derive/src/lib.rs
debris: I partly agree and I used to omit `new()` when struct had `default()`, but on the other hand `new()` and `default()` coexist in many other libraries (and even std), so I changed my mind and decided to have both of them, cause it makes using the library easier

derive/src/lib.rs
debris: That's not a runtime error. `include_bytes` include the contract at compile time.

Also, we already ensured that the file exists few lines before: https://github.com/paritytech/ethabi/pull/42/files/47784de909307c8279e065f5055cb35352a4a7c5#diff-11de442f8c13daab760fd46822928e3dR30

The only way to avoid loading the contract twice at compile time is to create `ethabi::Function` and `ethabi::Event` functions from scratch at compile time, and I wanted to avoid that in the first version of the library, cause it's a bit of a hassle

derive/src/lib.rs
debris: It's impossible to write a proof here. It's also impossible to propagate an error upstream

derive/src/lib.rs
debris: Agreed

tomusdrw:
Can't we review `ethabi_derive` crate separately then?

derive/src/lib.rs
tomusdrw: Oh, sorry I didn't notice it's already using `include_bytes`. So I suggest replacing `unwrap()` with `expect("JSON file is read during compilation and the same JSON file is included in the binary hence the value is available.")`
Even if we know that it cannot fail having different proofs like that helps reading and reasoning about the code.

derive/src/lib.rs
tomusdrw: Out of curiosity do you have an example of no-parameter `new()` in `std` for a structure that always have `::default()`?
Afair it's only used in cases when calling `::default()` would lead to a default generic parameter not being inferred properly.

ethabi/src/filter.rs
debris: done

tomusdrw:
Major version bump required, because it touches a lot of public methods.

debris:
yes! I will bump the version after derive is merged ;)

derive/src/lib.rs
debris: https://doc.rust-lang.org/std/collections/hash_map/struct.DefaultHasher.html

PR #43, opened 2017-08-22
use_contract! macro

debris:


derive/src/lib.rs
```diff
@@ -0,0 +1,513 @@
...
+use std::{env, fs};
+use std::path::PathBuf;
+use proc_macro::TokenStream;
+use heck::{SnakeCase, CamelCase};
+use ethabi::{Result, ResultExt, Contract, Event, Function, ParamType, Constructor};
+
+#[proc_macro_derive(EthabiContract, attributes(ethabi_contract_options))]
+pub fn ethabi_derive(input: TokenStream) -> TokenStream {
+	let s = input.to_string();
+	let ast = syn::parse_derive_input(&s).unwrap();
```
tomusdrw: why `unwrap`?

derive/src/lib.rs
```diff
@@ -0,0 +1,513 @@
...
+	let gen = impl_ethabi_derive(&ast).expect("`derive(EthabiContract)` failed");
+	gen.parse().unwrap()
```
tomusdrw: why `unwrap`?

derive/src/lib.rs
```diff
@@ -0,0 +1,513 @@
...
+					ethabi::Token::FixedArray(v)
+				}
+			}
+		},
+	}
+}
+
+fn from_token(kind: &ParamType, token: &syn::Ident) -> quote::Tokens {
+	match *kind {
+		ParamType::Address => quote! { #token.to_address().unwrap() },
```
tomusdrw: why `unwrap()`?

derive/src/lib.rs
```diff
@@ -0,0 +1,513 @@
...
+	}
+}
+
+fn impl_contract_event(event: &Event) -> quote::Tokens {
+	let query_name = string_ident(&event.name);
+	let name = syn::Ident::new(event.name.to_snake_case());
+	let event_name = syn::Ident::new(event.name.to_camel_case());
+	quote! {
+		pub fn #name(&self) -> events::#event_name {
+			events::#event_name::from_x(self.contract.event(#query_name).unwrap())
```
tomusdrw: Why `unwrap`?

derive/src/lib.rs
```diff
@@ -0,0 +1,513 @@
...
+			syn::Ident::new(format!("param{}", index))
+		} else {
+			param.name.to_snake_case().into()
+		}).collect();
+	let kinds: Vec<_> = constructor.inputs
+		.iter()
+		.map(|param| rust_type(&param.kind))
+		.collect();
+	let params: Vec<_> = names.iter().zip(kinds.iter())
+		.map(|(param_name, kind)| quote! { #param_name: #kind })
```
tomusdrw: Would be nice to have `Into<#kind>` if possible.

derive/src/lib.rs
```diff
@@ -0,0 +1,513 @@
...
+		})
+		.collect();
+
+	let topic_kinds: Vec<_> = event.inputs
+		.iter()
+		.filter(|param| param.indexed)
+		.map(|param| rust_type(&param.kind))
+		.collect();
+
+	let params: Vec<_> = topic_names.iter().zip(topic_kinds.iter())
```
tomusdrw: `Into<ethabi::Topic<#kind>>` if possible would be cool.

derive/src/lib.rs
```diff
@@ -0,0 +1,513 @@
...
+		}
+
+		impl Default for #name {
+			fn default() -> Self {
+				#name::new()
+			}
+		}
+
+		impl #name {
+			pub fn new() -> Self {
```
tomusdrw: The question from previous PR still applies :)

tests/src/lib.rs
```diff
@@ -0,0 +1,30 @@
+extern crate rustc_hex;
+extern crate ethabi;
+#[macro_use]
+extern crate ethabi_derive;
+#[macro_use]
+extern crate ethabi_contract;
+
+use_contract!(eip20, "Eip20", "../examples/eip20.json");
```
tomusdrw: Why is `use_contract!` macro necessary? Why not making it explicit:
```rust
pub mod eip20 {
  #[derive(EthabiContract)]
  #[ethabi_contract_options(path = $path)]
  struct Eip20;
}
```
looks less magical to me.

debris: It's how macros are done in `diesel` and there are few reasons for that:

1. people will write

    ```rust
    pub mod eip20 {
      #[derive(EthabiContract)]
      struct Eip20;
    }
    ```

    omitting `ethabi_contract_options`. The compilation will fail during compile time, instead of failing during syntax parsing

2. people will write

    ```rust
    pub mod contracts {
      #[derive(EthabiContract)]
      #[ethabi_contract_options(path = $path)]
      struct Eip20;

      #[derive(EthabiContract)]
      #[ethabi_contract_options(path = $path)]
      struct MyContract;
    }
    ```

    and it will fail, cause `EthabiContract` macro declares modules `logs`, `events` and `functions`

3. current implementation requires `struct Eip20` to load contract internally, so it needs `Contract` property. It's not possible to add fields to already declared struct

derive/src/lib.rs
debris: every single framework does that.

https://github.com/diesel-rs/diesel/blob/master/diesel_codegen/src/lib.rs#L97

it's not expected to fail. ever. It cannot be handled. I can add an exception here

derive/src/lib.rs
debris: it's not expected to fail. ever

tests/src/lib.rs
tomusdrw: Fair enough :)

derive/src/lib.rs
tomusdrw: It got answered, actually the simples case is `Vec::new()` and `Vec::default()`

derive/src/lib.rs
debris: done

derive/src/lib.rs
debris: done

PR #58, opened 2017-10-16
Remove parsing of ABI at runtime, closes #54

### LINKED ISSUE #54: ethabi_derive should not parse .abi file at runtime
the runtime parsing happens here: 

https://github.com/paritytech/ethabi/blob/75c2d9b18d61dcc41b0dcbea137ba62e36984713/derive/src/lib.rs#L119-L126

axelchalon:


derive/src/lib.rs
```diff
@@ -137,9 +130,11 @@ fn impl_ethabi_derive(ast: &syn::DeriveInput) -> Result<quote::Tokens> {
 	Ok(result)
 }
 
+/*
 fn string_ident(s: &str) -> syn::Ident {
 	syn::Ident::new(format!("\"{}\"", s))
 }
+*/
```
debris: remove this function

derive/src/lib.rs
```diff
@@ -350,14 +363,22 @@ fn impl_contract_constructor(constructor: &Constructor) -> quote::Tokens {
...
-				.as_ref()
-				.expect(INTERNAL_ERR)
-				.encode_input(code, &v)
-				.expect("encode_input not to fail; ethabi_derive bug")
+
+			ethabi::Constructor {
+				inputs: #constructor_inputs
+			}
+			.encode_input(code, &v)
+			.expect("encode_input not to fail; ethabi_derive bug")
```
debris: we could actually use `INTERNAL_ERR`

PR #66, opened 2017-11-13
Implement call_async

### LINKED ISSUE #64: Add optional call variant which uses futures
related https://github.com/paritytech/ethabi/pull/62#issuecomment-341708679

axelchalon:
closes https://github.com/paritytech/ethabi/issues/64

Added another macro `use_contract_futures!()` that derives the contract with the extra `call_future` function. Refactored the code a little bit to separate macro config parsing from the implementation; and to allow the macro config to be shared between the functions.

@tomusdrw's comment mentioned returning `IntoFuture<>`; currently the method returns a `Future`. Let me know if I should change this.

Also, I imagine we could write integration tests with rust-web3?

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -22,12 +22,11 @@ pub fn ethabi_derive(input: TokenStream) -> TokenStream {
 
 	let options = get_options(&ast.attrs, "ethabi_contract_options").expect(ERROR_MSG);
 	let futures = option_exists(&options, "futures");
-	let path = get_option_value(&options, "path").expect(ERROR_MSG);
-	let name = get_option_value(&options, "name").expect(ERROR_MSG);
+	let path = option_value(&options, "path").expect(ERROR_MSG);
+	let name = option_value(&options, "name").expect(ERROR_MSG);
 	let implementer = EthabiDeriveImplementer { futures, path, name };
 
 	let gen = implementer.implement().expect(ERROR_MSG);
-
 	gen.parse().expect(ERROR_MSG)
 }
 
@@ -49,7 +48,7 @@ fn option_exists<'a>(options: &'a [syn::MetaItem], name: &str) -> bool {
 	options.iter().find(|a| a.name() == name).is_some()
 }
 
-fn get_option_value<'a>(options: &'a [syn::MetaItem], name: &str) -> Result<&'a str> {
+fn option_value<'a>(options: &'a [syn::MetaItem], name: &str) -> Result<&'a str> {
 	let item = options.iter().find(|a| a.name() == name).chain_err(|| format!("Expected to find option {}", name))?;
 	value_of_meta_item(item, name)
 }
@@ -494,7 +493,7 @@ impl<'a> EthabiDeriveImplementer<'a> {
 
 			let call_future_quote = if self.futures {
 				quote! {
-					pub fn call_future</*'s,*/#(#template_params),*>(/*'s */self, #(#params ,)* do_call: &Fn(ethabi::Bytes) -> Box<Future<Item=ethabi::Bytes, Error=String>>) -> Box<Future<Item=#output_kinds, Error=ethabi::Error> /*+ 's*/>
+					pub fn call_future<#(#template_params),*>(self, #(#params ,)* do_call: &Fn(ethabi::Bytes) -> Box<Future<Item=ethabi::Bytes, Error=String>>) -> Box<Future<Item=#output_kinds, Error=ethabi::Error>>
 					{
 						let encoded_input = self.input(#(#names),*);
 
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -21,10 +21,9 @@ pub fn ethabi_derive(input: TokenStream) -> TokenStream {
 	let ast = syn::parse_derive_input(&s).expect(ERROR_MSG);
 
 	let options = get_options(&ast.attrs, "ethabi_contract_options").expect(ERROR_MSG);
-	let async = option_exists(&options, "async");
 	let path = option_value(&options, "path").expect(ERROR_MSG);
 	let name = option_value(&options, "name").expect(ERROR_MSG);
-	let implementer = EthabiDeriveImplementer { async, path, name };
+	let implementer = EthabiDeriveImplementer { path, name };
 
 	let gen = implementer.implement().expect(ERROR_MSG);
 	gen.parse().expect(ERROR_MSG)
@@ -44,10 +43,6 @@ fn get_options(attrs: &[syn::Attribute], name: &str) -> Result<Vec<syn::MetaItem
 	}
 }
 
-fn option_exists<'a>(options: &'a [syn::MetaItem], name: &str) -> bool {
-	options.iter().find(|a| a.name() == name).is_some()
-}
-
 fn option_value<'a>(options: &'a [syn::MetaItem], name: &str) -> Result<&'a str> {
 	let item = options.iter().find(|a| a.name() == name).chain_err(|| format!("Expected to find option {}", name))?;
 	value_of_meta_item(item, name)
@@ -61,7 +56,6 @@ fn value_of_meta_item<'a>(item: &'a syn::MetaItem, name: &str) -> Result<&'a str
 }
 
 struct EthabiDeriveImplementer<'a> {
-	async: bool,
 	path: &'a str,
 	name: &'a str,
 }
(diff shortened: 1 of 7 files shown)


derive/src/lib.rs
```diff
@@ -144,530 +44,675 @@ fn get_options(attrs: &[syn::Attribute], name: &str) -> Result<Vec<syn::MetaItem
...
-		let o_impl = match function.outputs.len() {
-			0 => quote! { Ok(()) },
-			1 => {
-				let o = "out".into();
-				let from_first = from_token(&function.outputs[0].kind, &o);
+			let call_future_quote = if self.futures {
 				quote! {
-					let out = self.function.decode_output(output)?.into_iter().next().expect(super::INTERNAL_ERR);
-					Ok(#from_first)
+					pub fn call_future<#(#template_params),*>(self, #(#params ,)* do_call: &Fn(ethabi::Bytes) -> Box<Future<Item=ethabi::Bytes, Error=String>>) -> Box<Future<Item=#output_kinds, Error=ethabi::Error>>
```
tomusdrw: We might get rid of one of the `Boxes` by doing this (it will also make user APIs more flexible).

```rust
pub fn call_async<__F, __T, __Fn, #(#template_params ,)*>(self, #(params ,)* do_call: __Fn)
  -> Box<Future<Item=#output_kinds, Error=ethabi::Error>>
where
  __F: Future<Item=ethabi::Bytes, Error=String>,
  __T: IntoFuture<Future=__F, Item=ethabi::Bytes, Error=String>,
  __Fn: FnOnce(ethabi::Bytes) -> __T
{

}
```

In future we might:
 - use  `impl Future` as the output type
 - Allow `Into<ethabi::Bytes>` to be returned from `__Fn`
 - Allow custom errors to be returned (`ethabi::Error` could just box them)

tests/src/lib.rs
```diff
@@ -127,5 +130,26 @@ mod tests {
...
+		use futures::{Future, future};
+
+		let contract = Eip20Futures::default();
+		let address_param = [0u8; 20];
+		let functions = contract.functions(); // .balance_of() is moved
+
+		let result = functions.balance_of().call_future(address_param, &|data| {
+			assert_eq!(data, "70a082310000000000000000000000000000000000000000000000000000000000000000".from_hex().unwrap());
+			Box::new(future::ok("000000000000000000000000000000000000000000000000000000000036455b".from_hex().unwrap()))
+        });
```
tomusdrw: Something funky with indents here.

tests/src/lib.rs
```diff
@@ -1,13 +1,16 @@
...
+extern crate futures;
+#[allow(unused_imports)]
 extern crate ethabi;
 #[macro_use]
 extern crate ethabi_derive;
 #[macro_use]
 extern crate ethabi_contract;
 
 use_contract!(eip20, "Eip20", "../res/eip20.abi");
+use_contract_futures!(eip20_futures, "Eip20Futures", "../res/eip20.abi");
```
tomusdrw: I think would be better to call it `async` instead of `futures` here as well.

rphmeier:
the async variant is a superset of the sync variant. `Result<T, E>` is `IntoFuture`, so if your function returns a Result you can use it just like `call` currently.

tomusdrw:
The difference is in the return type. `call` should pretty much be implemented as;
```rust
pub fn call(/* macro mumbo-jumbo*/) -> Result<..> {
   self.call_async(...).wait()
} 
```

The provider of `FnOnce(ethabi::Bytes) -> Result<ethabi::Bytes, String>` function indeed doesn't really care which of the functions the user of `ethabi` decides to use.

rphmeier:
if having support for both `call` and `call_async` introduces a large amount of complexity in the implementation, why not just recommend that users call `wait()` on the produced future if they want to

axelchalon:
@debris also mentioned that users might not want to have to import Futures if they simply want to use the synchronous call variant, hence having the procedural macro attribute for enabling/disabling the `futures` feature and having two different functions.

tomusdrw:
IMHO there should be only one macro that produces two methods:
```rust
pub fn call(self, f: Fn(Bytes) -> Result<Bytes, String>) -> Result<_, _> {
  $crate::futures::wait(self.call_async(f))
}
pub fn call_async(self, f: Fn(Bytes) -> IntoFuture<...>) -> Box<Future<..>> {
  // Do some magic stuff.
}
```

This creates no code duplication, no need to import futures and the user of the library is able to choose which method he wants.

axelchalon:
hey, any consensus on this? I'm halting the refactoring of parity using ethabi-derive because I'm afraid the syntax is going to change and I'll have to go back on my PRs
@tomusdrw but if the macro always defines this call_async function, you'll still need to add the futures as dependency where you'll use the macro, no?

tomusdrw:
No, futures should only be a dependency of `ethabi` (actually can be even hidden under a rust feature).
You can refer using `$crate::futures` to the dependencies of your crate inside the macro.

tomusdrw:
@debris @rphmeier what are you thoughts on it?

derive/src/lib.rs
```diff
@@ -144,530 +44,681 @@ fn get_options(attrs: &[syn::Attribute], name: &str) -> Result<Vec<syn::MetaItem
...
+	async: bool,
+	path: &'a str,
+	name: &'a str,
 }
 
-fn impl_contract_function(function: &Function) -> quote::Tokens {
-	let name = syn::Ident::new(function.name.to_snake_case());
-	let function_name = syn::Ident::new(function.name.to_camel_case());
+impl<'a> EthabiDeriveImplementer<'a> {
+	pub fn implement(self: &Self) -> Result<quote::Tokens> {
```
debris: why not just `&self`?

derive/src/lib.rs
```diff
@@ -19,115 +19,15 @@ const ERROR_MSG: &'static str = "`derive(EthabiContract)` failed";
...
+	let async = option_exists(&options, "async");
+	let path = option_value(&options, "path").expect(ERROR_MSG);
+	let name = option_value(&options, "name").expect(ERROR_MSG);
+	let implementer = EthabiDeriveImplementer { async, path, name };
 
-		#functions_quote
-	};
-
-	Ok(result)
+	let gen = implementer.implement().expect(ERROR_MSG);
```
debris: there is no need to create a structure if this structure has only one function. Standalone function is better

derive/src/lib.rs
```diff
@@ -144,530 +44,681 @@ fn get_options(attrs: &[syn::Attribute], name: &str) -> Result<Vec<syn::MetaItem
...
-			let inner: syn::Ident = "inner".into();
-			let inner_loop = from_token(kind, &inner);
+		};
+
+		let futures_import_quote = if self.async {
 			quote! {
-				#token.to_array().expect(super::INTERNAL_ERR).into_iter()
-					.map(|#inner| #inner_loop)
-					.collect()
+				extern crate futures;
```
debris: @tomusdrw is right, you could refer to a `futures` crate using `$crate::futures`. Then you could also add an optional `feature` to `contract/Cargo.toml`.

```toml
[dependencies]
futures = { version = "0.1", optional = true }

[features]
async = ["futures"]
```

derive/src/lib.rs
axelchalon: the structure has one public function but several other functions, and they each need to access the struct fields

derive/src/lib.rs
axelchalon: $crate doesn't seem to work within the procedural macro (https://stackoverflow.com/questions/44950574/using-crate-in-rusts-procedural-macros also) ; gives me:

```
error: unknown macro variable `crate`                                                                                                                                    
  --> tests/src/lib.rs:11:1                                                                                                                                              
   |                                                                                                                                                                     
11 | use_contract!(eip20, "Eip20", "../res/eip20.abi");                                                                                                                  
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                  
   |                                                                                                                                                                     
   = note: this error originates in a macro outside of the current crate
```

A workaround would be to use `use ::ethabi_contract::futures::{Future, IntoFuture};` instead (inside the proc macro), but then it means we cannot import the contract crate with another name

tomusdrw: We have that limitation for `ethabi` already, so imho doesn't really matter.

axelchalon: Ok!

Hmm if I understood correctly, if we make an `async` feature then we cannot rely on `call` calling `.wait()` on `call_async` (i.e. not possible to implement it like https://github.com/paritytech/ethabi/pull/66#issuecomment-344226556)

axelchalon:
I agree with Rob that we should only keep `call_async()` and remove `call()`. The callback to `call_async()` has to return an `IntoFuture` so it can return a `Future` or a `Result` for example, which is convenient enough. `call_async()` returns a future; we can simply `.wait()` on it if we want to. It's too redundant to have both.

On another note, Tomek did some nice work on his fork https://github.com/tomusdrw/ethabi 
 and we should merge it into this branch/PR some time soon.

PR #71, opened 2017-12-05
Fix constructor input encoding

lght:


debris:
ahh, it was an ugly bug :/ would be good to also get tests for this

lght:
> ahh, it was an ugly bug :/ would be good to also get tests for this

agreed, spent quite some time with @tomusdrw tracking this down. know that we're planning / working on changing a lot of ethabi's API and type structure. 

will likely wait for those changes, unless you have suggestions for writing general tests that won't have to be refactored too much?

PR #80, opened 2018-02-02
Fix lenient encoding of big uint256 (pending: i256)

### LINKED ISSUE #2: Fix lenient parsing of negative integers.
Basically make [this test](https://github.com/ethcore/ethabi/blob/master/src/bin/main.rs#L217-L224) work.

### LINKED ISSUE #79: uint256 encoding problem
When I am trying to encode big uint256 value I got this error

```
Cannot parse uint256

Caused by:
  number too large to fit in target type
```

I'am trying to do 
```
ethabi encode function abi.json mint -p f3eb3b7d01416d22fdc051c313af5634afac316e -p 10000000000 --lenient
```
or simply
```
ethabi encode params -v uint256 10000000000 --lenient
```
If I trying to encode smaller number like 1000 it works fine (up to 10 ** 9) but starting from 10 ** 10 it raises error. Why? uint256 can store values up to 2 ** 256 - 1 as I understand 

Is it a bug or I am using ethabi incorrectly?

axelchalon:
should fix #79

(Previous behaviour was to parse the input u256/i256 string as u32/i32 and then pad it to 256 bits; hence why it didn't work with big numbers as input. Now I'm trying to parse the input string as u256/i256 directly.)

- Fix lenient encoding of big u256
- Add test for lenient encoding of big u256
- Add test for lenient encoding of big i256. Didn't pass before; still doesn't pass. It looks like ethabi::Int and ethabi::Uint are both aliases to ethereum_types::U256 and so I don't know how to handle converting a string to a signed i256. Do we have anything in ethereum-types to convert from a string to a signed i256? Or is there a util function somewhere I could use? Any pointers? @debris 
- Un-ignore test for encoding negative i256 (I think we can close #2)

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -214,12 +214,28 @@ mod tests {
 		assert_eq!(execute(command).unwrap(), expected);
 	}
 
-	// TODO: parsing negative values is not working
 	#[test]
-	#[ignore]
 	fn int_encode() {
-		let command = "ethabi encode paramas -v int256 -2 --lenient".split(" ");
+		let command = "ethabi encode params -v int256 100000000000 --lenient".split(" ");
+		let expected = "000000000000000000000000000000000000000000000000000000174876e800";
+		assert_eq!(execute(command).unwrap(), expected);
+	}
+
+	#[test]
+	fn int_encode_negative() {
+		let command = "ethabi encode params --lenient -v -- int256 -2".split(" ");
 		let expected = "fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe";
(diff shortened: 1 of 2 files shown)


snd:
how can i help?

dvdplm:
@axelchalon currently `ethereum-types` does not implement an `I256`. I think it'd be useful to make a new PR with the `U256` fixes here and get that merged. Consider opening an issue on the `uint` crate [here](https://github.com/paritytech/parity-common/issues) for signed bigint support.

tomusdrw:
This looks extremely stale, please re-open in case we still need this.

PR #103, opened 2018-06-17
Build empty constructor when none specified

udoprog:
Hey!

This changes the derive implementation to always include an empty constructor, even if none is available in the ABI specification.

This makes it easier to use a uniform deploy implementation, even when there are no constructors specified in the solidity contract.


debris:
thanks!

PR #107, opened 2018-06-20
Simplify derive to remove unused wrapper structs

udoprog:
This gets rid of the struct that currently doesn't do anything.

You can see the difference in `tests/src/lib.rs`. There's no need to store a local variable called `contract`.

This also simplifies the `use_contract` macro, which no longer needs a `name` variable. All though I've not incorporated this change here since we use [our own macro](https://github.com/PrimaBlock/parables/blob/master/testing/macros.rs) in parables.

For me, this makes tests cleaner.

### DIFF
--- a/contract/src/lib.rs
+++ b/contract/src/lib.rs
@@ -1,6 +1,6 @@
 #[macro_export]
 macro_rules! use_contract {
-	($module: ident, $name: expr, $path: expr) => {
+	($module: ident, $path: expr) => {
 		#[allow(dead_code)]
 		#[allow(missing_docs)]
 		#[allow(unused_imports)]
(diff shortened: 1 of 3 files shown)


udoprog:
Thanks, I added an update to `use_contract` as well now.

udoprog:
Branch rebased to take #105 into account.

debris:
makes sense! in the past, structure was used to load abi in the runtime, but this is not a case any more

dvdplm:
Q: I was upgrading `parity-ethereum` to use the latest `ethabi` (v5 => v6) and there are a bunch of `use_contract!` invocations that pass the `$name`. Without it however, there are many errors like this:

```
767 |             registrar: registry::Registry::default(),
    |                                  ^^^^^^^^ Could not find `Registry` in `registry`
```

So I wonder, what is the recommended way to fix this? How should consuming code access the contract?

udoprog:
@dvdplm IIRC:

Functions are available under `registry::functions::<name-of-function>`
Events are available under `registry::events::<name-of-event>`.
Logs are available under `registry::logs::<name-of-log>`.

So if you previously did:

```rust
let registry = registry::Registry::default();
registry.foo()
```

You would now do:

```rust
registry::functions::foo()
```

What I do for convenience is to pull the functions module into scope:

```rust
use registry::functions as registry;
registry::foo();
```

PR #108, opened 2018-06-20
Introduce LogFilter trait to simplify filtering APIs

udoprog:
This simplifies writing APIs that uses the log abstraction and just wants "everything" without filtering.

You can see an example of it here:
https://github.com/PrimaBlock/parables/commit/0410483fa901d506273572edcd3ae3a76a3e4e0b#diff-1e74411566a577ec5675e59bf0bff74fR68

There's no longer a need to call `create_filter` with a ton of `Topic::Any` (or `None`). Instead we can generically make use of `LogFilter::match_any`.

### DIFF
--- a/ethabi/src/log.rs
+++ b/ethabi/src/log.rs
@@ -1,4 +1,10 @@
-use {Hash, Token, Bytes, Result};
+use {Hash, Token, Bytes, Result, TopicFilter};
+
+/// Common filtering functions that are available for any event.
+pub trait LogFilter {
+	/// Match any log parameters.
+	fn match_any(&self) -> TopicFilter;
+}
 
 /// trait common to things (events) that have an associated `Log` type
 /// that can be parsed from a `RawLog`
--- a/ethabi/src/log.rs
+++ b/ethabi/src/log.rs
@@ -3,7 +3,7 @@ use {Hash, Token, Bytes, Result, TopicFilter};
 /// Common filtering functions that are available for any event.
 pub trait LogFilter {
 	/// Match any log parameters.
-	fn match_any(&self) -> TopicFilter;
+	fn wildcard_filter(&self) -> TopicFilter;
 }
 
 /// trait common to things (events) that have an associated `Log` type
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -47,6 +47,7 @@ fn impl_ethabi_derive(ast: &syn::DeriveInput) -> Result<quote::Tokens> {
 			pub mod events {
 				use ethabi;
 				use ethabi::ParseLog;
+				use ethabi::LogFilter;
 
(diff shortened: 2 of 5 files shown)


udoprog:
Should be good now, thanks!

ethabi/src/log.rs
```diff
@@ -1,4 +1,10 @@
-use {Hash, Token, Bytes, Result};
+use {Hash, Token, Bytes, Result, TopicFilter};
+
+/// Common filtering functions that are available for any event.
+pub trait LogFilter {
```
debris: I can't see how having a trait is useful here

ethabi/src/log.rs
```diff
@@ -1,4 +1,10 @@
...
+	/// Match any log parameters.
+	fn match_any(&self) -> TopicFilter;
```
debris: I find this function name confusing. We are not matching anything here. Just building a TopicFilter that matches events of `self` kind.

ethabi/src/log.rs
udoprog: This allows us to generically convert generated filter types into a `TopicFilter` that matches anything (https://github.com/PrimaBlock/parables/blob/master/testing/evm.rs#L524) which can improve ergonomics for the cases where you just want to match a given log type (https://github.com/PrimaBlock/parables/blob/master/example/src/main.rs#L56 - `with_filter` is optional here).

ethabi/src/log.rs
udoprog: Sure, any suggestions on what I should name it? The generated one is called `create_filter`.

ethabi/src/log.rs
debris: I know :) I just wonder why do we need to use a trait, when a simple function is enough :)

ethabi/src/log.rs
debris: naming is always difficult. maybe `create_empty_filter` or `create_filter_for_all`. @tomusdrw what do you think?

ethabi/src/log.rs
tomusdrw: @debris that's the whole point, that using a trait allows you to write generic code that handles all possible events. You can't do that with simple function, unless your point is that because there is nothing else in that particular trait the caller can just call that one function and pass `TopicFilter` instead of `F: LogFilter`?

ethabi/src/log.rs
tomusdrw: maybe `create_wildcard_filter`? TBH I would drop `create_` from both this function and `create_filter`.

debris: so `filter` and `wildcard_filter` it is

udoprog: Rebased and renamed. Thanks! <3

PR #112, opened 2018-07-02
improve error message when trying to decode empty bytes

snd:
as discussed in bridge riot. jsonrpc responds with result `0x` in case contract or method don't exist. ethabi fails with `Cannot decode {type}` in that case which suggests a problem different from the actual problem.

did experiments and found out that the only ethabi types that can be decoded from empty bytes are the "nonsense" types `FixedArray` and `FixedBytes` with length `0`

### DIFF
--- a/ethabi/src/decoder.rs
+++ b/ethabi/src/decoder.rs
@@ -36,6 +36,10 @@ fn as_bool(slice: &[u8; 32]) -> Result<bool, Error> {
 
 /// Decodes ABI compliant vector of bytes into vector of tokens described by types param.
 pub fn decode(types: &[ParamType], data: &[u8]) -> Result<Vec<Token>, Error> {
+    let is_empty_bytes_valid_encoding = types.iter().all(|t| t.is_empty_bytes_valid_encoding());
+    if !is_empty_bytes_valid_encoding && data.len() == 0 {
+        bail!("please ensure the contract and method you're calling exist! failed to decode empty bytes. if you're using jsonrpc this is likely due to jsonrpc returning `0x` in case contract or method don't exist");
+    }
 	let slices = slice_data(data)?;
 	let mut tokens = vec![];
 	let mut offset = 0;
--- a/ethabi/src/decoder.rs
+++ b/ethabi/src/decoder.rs
@@ -479,5 +479,25 @@ mod tests {
 		let decoded = decode(&[ParamType::String], &encoded).unwrap();
 		assert_eq!(decoded, expected);
 	}
+
+	#[test]
+	fn decode_from_empty_byte_slice() {
+        // these can NOT be decoded from empty byte slice
+        assert!(decode(&[ParamType::Address], &[]).is_err());
+        assert!(decode(&[ParamType::Bytes], &[]).is_err());
+        assert!(decode(&[ParamType::Int(0)], &[]).is_err());
+        assert!(decode(&[ParamType::Int(1)], &[]).is_err());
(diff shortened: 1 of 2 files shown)


ethabi/src/decoder.rs
```diff
@@ -36,6 +36,10 @@ fn as_bool(slice: &[u8; 32]) -> Result<bool, Error> {
 
 /// Decodes ABI compliant vector of bytes into vector of tokens described by types param.
 pub fn decode(types: &[ParamType], data: &[u8]) -> Result<Vec<Token>, Error> {
+    let is_empty_bytes_valid_encoding = types.iter().all(|t| t.is_empty_bytes_valid_encoding());
```
tomusdrw: I wonder if this check is needed (thinking if it could be simplified to `types.len() == 1` && types[0].is_empty)).

What's the return value for type `(Bytes, Bytes)`? if both are empty? is it really `0x` or should it actually be encoded somehow that it's a tuple?

snd: > I wonder if this check is needed (thinking if it could be simplified to types.len() == 1 && types[0].is_empty)).

it could be that there are multiple ParamTypes in `types` that can be "empty".

> What's the return value for type (Bytes, Bytes)? if both are empty? is it really 0x or should it actually be encoded somehow that it's a tuple?

you mean for encoding? is a tuple a valid `ParamType`? btw: it seems to me that a `Bytes` can never be decoded from empty bytes since it's always lead by it's length

tomusdrw: So correct me if I'm wrong, but my understanding was that this function takes two parameters that mean:
1. `types` - a slice of returned `ParamTypes` - i.e. if solidity function returns `(Bytes, Bytes)` this is going to be `&[Bytes, Bytes]` 
2. `data` the binary encoding of returned data that matches the `types`, the data is then `sliced` and decoded to correct `ParamTypes`.

So my gut feeling is that the only case when you actually get `0x` as a valid response from solidity is one the function:
1. Returns only one result (i.e. `(Bytes)` or `(Bytes[])`)
2. It's empty bytes or empty array

Because of this I think the check could be simplified to:
```
let is_empty_result_allowed = types.len() == 1 && types[0].is_bytes_or_empty_array();
if data.len() == 0 && !is_empty_result_allowed {
  bail!()
}
```
But I'm not solidity/ethabi expert, so best to wait for @debris opinion.

snd:
@debris are `FixedBytes` and `FixedArray` with len `0` even allowed? imho they make no sense. if we'd disallow them we'd only have to check for `data.len == 0` since they are the only types which can be decoded from "nothing".

ethabi/src/decoder.rs
snd: > But I'm not solidity/ethabi expert, so best to wait for @debris opinion.

me neither. what do you think @debris

snd:
i've added a test that checks my assumptions on what can be decoded from empty byte slices.

again note that `String` and `Bytes` can NOT be decoded from empty byte slices. they are prefixed by their length.

snd:
@debris ping

ethabi/src/decoder.rs
debris: > it could be that there are multiple ParamTypes in types that can be "empty".

I think this is true

> types - a slice of returned ParamTypes - i.e. if solidity function returns `(Bytes, Bytes)` this is going to be `&[Bytes, Bytes]`

iirc, it's will be encoded as two independent elements appended one after another. I believe @snd is right here

PR #120, opened 2018-08-28
Fix v5.1

dvdplm:
Current v5.1 does not work with `parity-ethereum`. This back-ports a critical bug fix for String params (#20eb865) and bumps the ethereum-types and rustc-hex dependencies.

This should be merged to the `v5.1.0` tag and published as 5.1.2 to unblock https://github.com/paritytech/parity-ethereum/pull/9410


debris:
I believe this pull request should be opened against a different branch than `master`

dvdplm:
> I believe this pull request should be opened against a different branch than master

Ooops, my bad. See https://github.com/paritytech/ethabi/pull/121

PR #121, opened 2018-08-28
Backport fixes and update deps at v5.1

dvdplm:
Instead of https://github.com/paritytech/ethabi/pull/120


dvdplm:
Thanks for the review. I will tag this `v5.1.3` and publish it to crates.

PR #125, opened 2018-09-13
update ethabi-derive dependencies

debris:
syn && quote to their latest versions

### DIFF
--- a/derive/src/constructor.rs
+++ b/derive/src/constructor.rs
@@ -1,4 +1,5 @@
-use {quote, ethabi};
+use ethabi;
+use proc_macro2::TokenStream;
 
 use super::{
 	input_names, template_param_type, rust_type, get_template_names, to_token, from_template_param,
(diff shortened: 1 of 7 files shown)


derive/src/constructor.rs
```diff
@@ -129,6 +130,6 @@ mod tests {
 			}
 		};
 
-		assert_eq!(expected, c.generate());
+		assert_eq!(expected.to_string(), c.generate().to_string());
```
debris: `TokenStream` does not implement `PartialEq`, so I convert it to string as a workaround

debris:
Once merged, I'll release `ethabi-derive 6.0.2`

derive/src/constructor.rs
dvdplm: Out of curiosity, is there (a good) reason why it doesn't?

debris: I guess someone just overlooked it during the refactor of `quote` and `proc-macro`

debris:
`6.0.2` released

PR #128, opened 2018-10-16
add `fn is_empty()` and implement `std::ops::Index` for Topic

### LINKED ISSUE #127: add `fn is_empty` and implement `std::ops::Index` for `struct Topic`
simplifies usage of the `Topic` struct. e.g. 

https://github.com/paritytech/parity-bridge/pull/200/files/82f6c649059002f92378761d0f990ec438583e65#diff-58a128532db808c036b80e23d4b2eb99

debris:
closes #127

### DIFF
--- a/ethabi/src/filter.rs
+++ b/ethabi/src/filter.rs
@@ -1,3 +1,4 @@
+use std::ops;
 use serde::{Serialize, Serializer};
 use serde_json::Value;
 use hex::ToHex;
@@ -52,6 +53,14 @@ impl<T> Topic<T> {
 			Topic::Any => Topic::Any,
 			Topic::OneOf(topics) => Topic::OneOf(topics.into_iter().map(f).collect()),
 			Topic::This(topic) => Topic::This(f(topic)),
+		}
+	}
+
+	/// Returns true if topic is empty (Topic::Any)
+	pub fn is_empty(&self) -> bool {
+		match *self {
+			Topic::Any => true,
+			Topic::This(_) | Topic::OneOf(_) => false,
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


ethabi/src/filter.rs
```diff
@@ -54,6 +55,14 @@ impl<T> Topic<T> {
 			Topic::This(topic) => Topic::This(f(topic)),
 		}
 	}
+
+	/// Returns true if topic is empty (Topic::Any)
+	pub fn is_empty(&self) -> bool {
+		match *self {
+			Topic::Any => true,
+			Topic::This(_) | Topic::OneOf(_) => false,
```
bkchr: Maybe for the second arm you could just use `_`?

tomusdrw: I like the fact that in case we extend `Topic` enum an implementer will be forced to explicitly handle the new case (so the compiler will notify him about this spot).
That said, I don't think it will happen in that particular case, so fine either way.

ethabi/src/filter.rs
```diff
@@ -54,6 +55,14 @@ impl<T> Topic<T> {
 			Topic::This(topic) => Topic::This(f(topic)),
 		}
 	}
+
+	/// Returns true if topic is empty (Topic::Any)
+	pub fn is_empty(&self) -> bool {
```
tomusdrw: Why not call it `is_any()`?

ethabi/src/filter.rs
```diff
@@ -111,6 +120,18 @@ impl Serialize for Topic<Hash> {
 	}
 }
 
+impl<T> ops::Index<usize> for Topic<T> {
+	type Output = T;
+
+	fn index(&self, index: usize) -> &Self::Output {
+		match *self {
+			Topic::Any => panic!("Topic is empty"),
+			Topic::This(ref topic) => topic,
```
tomusdrw: Should we panic if `index != 0` in case of `This`?

ethabi/src/filter.rs
debris: > I like the fact that in case we extend Topic enum an implementer will be forced to explicitly handle the new case

yes, I am a fan of defensive programming ;)

ethabi/src/filter.rs
debris: you are absolutely right! I will add a test case! :)

PR #140, opened 2018-11-14
Fix decoder to handle data with a length that is not a multiple of 32

### LINKED ISSUE #139: Decoding event data broken if data length is not a multiple of 32
## Problem description

Livepeer's `JobsManager` contract emits `NewJob(address,uint256,string,string,uint256,uint256)` events. Most (if not all) if these events have a data length that is not a multiple of 32. Here are two examples:

* https://etherscan.io/tx/0x712ed897c42e20de0f369eaaea32acc85fa9dd5f88e93170f64e00c9351cbc4a#eventlog
* https://etherscan.io/tx/0xc8d552532338e9ee470c1912d42c2fda0b41bdc1d8aebd5831ff43abcf6a24fd#eventlog

Trying to decode the event params with e.g.
```rust
let encoded = hex!(
    "
     0000000000000000000000000000000000000000000000000000000000000000
     00000000000000000000000000000000000000000000000000000000000000a0
     0000000000000000000000000000000000000000000000000000000000000152
     0000000000000000000000000000000000000000000000000000000000000001
     000000000000000000000000000000000000000000000000000000000054840d
     0000000000000000000000000000000000000000000000000000000000000092
     3132323033393637623533326130633134633938306235616566666231373034
     3862646661656632633239336139353039663038656233633662306635663866
     3039343265376239636337366361353163636132366365353436393230343438
     6533303866646136383730623565326165313261323430396439343264653432
     3831313350373230703330667073313678390000000000000000000000000000
     0000000000000000000000000000000000103933633731376537633061363531
     3761
"
);

let decoded = decode(
    &[
        ParamType::Uint(256),
        ParamType::String,
        ParamType::String,
        ParamType::Uint(256),
        ParamType::Uint(256),
    ],
    &encoded,
).unwrap()
```
currently fails because `util::slice_data` expects the length of the encoded data to be  multiple of 32.

## Web3.js behavior

Web3.js parses the same data fine, as the following log from the Node.js shell demonstrates:
```js
> const Web3 = require('web3');
> const web3 = new Web3();
> web3.eth.abi.decodeParameters(['uint256', 'string', 'string', 'uint256', 'uint256'], '0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000001520000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000054840d00000000000000000000000000000000000000000000000000000000000000923132323033393637623533326130633134633938306235616566666231373034386264666165663263323933613935303966303865623363366230663566386630393432653762396363373663613531636361323663653534363932303434386533303866646136383730623565326165313261323430396439343264653432383131335037323070333066707331367839000000000000000000000000000000000000000000000000000000000000001039336337313765376330613635313761')
Result {
  '0': '0',
  '1':
   '12203967b532a0c14c980b5aeffb17048bdfaef2c293a9509f08eb3c6b0f5f8f0942e7b9cc76ca51cca26ce546920448e308fda6870b5e2ae12a2409d942de428113P720p30fps16x9',
  '2': '93c717e7c0a6517a',
  '3': '1',
  '4': '5538829',
  __length__: 5 }
```

## Desired behavior

The desired behavior would be for `ethabi` to be able to decide the event data in the same way `web3.js` does it.

Jannis:
Fixes #139. The following changes are made:

1. The assumption that the length of encoded data is always a multiple of 32 is removed.
2. The encoded data is no longer sliced into 32-byte arrays. Instead, the decoder operates on the original data (a flat array of bytes).
3. A test is added to verify that decoding events such as the ones mentioned in #139 are decoded correctly.

This is inspired by the ABI decoder used in web3.js, specifically https://github.com/ethers-io/ethers.js/blob/master/src.ts/utils/abi-coder.ts#L1064.

### DIFF
--- a/ethabi/src/decoder.rs
+++ b/ethabi/src/decoder.rs
@@ -525,4 +525,35 @@ mod tests {
 			]
 		);
 	}
+
+	#[test]
+	fn decode_after_fixed_bytes_with_less_than_32_bytes() {
+		let encoded = hex!("
+			0000000000000000000000008497afefdc5ac170a664a231f6efb25526ef813f
+			0000000000000000000000000000000000000000000000000000000000000000
+			0000000000000000000000000000000000000000000000000000000000000000
+			0000000000000000000000000000000000000000000000000000000000000080
+			000000000000000000000000000000000000000000000000000000000000000a
+			3078303030303030314600000000000000000000000000000000000000000000
+		");
+
+		assert_eq!(
+			decode(
+				&[
+					ParamType::Address,
+					ParamType::FixedBytes(32),
+					ParamType::FixedBytes(4),
+					ParamType::String,
+				],
+				&encoded,
+			)
+			.unwrap(),
+			&[
+				Token::Address(hex!("8497afefdc5ac170a664a231f6efb25526ef813f").into()),
+				Token::FixedBytes([0u8; 32].to_vec()),
+				Token::FixedBytes([0u8; 4].to_vec()),
+				Token::String("0x0000001F".into()),
+			]
+		);
+	}
 }
--- a/ethabi/src/decoder.rs
+++ b/ethabi/src/decoder.rs
@@ -115,6 +115,8 @@ fn decode_param(param: &ParamType, data: &[u8], offset: usize) -> Result<DecodeR
 			Ok(result)
 		}
 		ParamType::FixedBytes(len) => {
+			// FixedBytes is anything from bytes1 to bytes32. These values
+			// are padded with trailing zeros to fill 32 bytes.
 			let bytes = take_bytes(data, offset, len)?;
 			let result = DecodeResult {
 				token: Token::FixedBytes(bytes),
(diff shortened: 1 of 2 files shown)


Jannis:
I'll look into why the test suite is failing.

Jannis:
It was a small oversight in advancing the decoder offset after a bool. That's fixed now.

Jannis:
@tomusdrw @debris Is there anything I can do to help? Anything that needs to be tested?

debris:
Hey, thanks for the pull request. I just came back from vacation and am trying to catch up with all issues / prs.

> The assumption that the length of encoded data is always a multiple of 32 is removed.

tbh. I don't understand why is it removed? What is the use case? When do we have to work with such data?

Jannis:
@debris See the issue this PR resolves: https://github.com/paritytech/ethabi/issues/139.

Jannis:
Question: Even if it is or was a bug in the Solidity compiler, there are contracts built with that version on mainnet today. Wouldn't it be a good idea to be robust enough to handle those situations?

tomusdrw:
Thank you for the PR @Jannis. We've decided not to support buggy ABIs in the library, see: #152 

> Question: Even if it is or was a bug in the Solidity compiler, there are contracts built with that version on mainnet today. Wouldn't it be a good idea to be robust enough to handle those situations?

We can't support every possible buggy ABI that is out there. Having a strict spec will benefit the entire ecosystem and make code of libraries in all languages simpler. To support old contracts you can always pre-process the ABI or the request/response returned by the contract to be aligned to 32-bytes.

PR #141, opened 2018-11-16
Fix encode array

driftluo:
In the encode array, everything after the single quote is ignored.:

```bash
$ ethabi encode params -v bool[] '[false"]'
00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000000
```

```bash
$ ethabi encode params -v bool[] '[1,false"]'
000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000001
```

```bash
$ ethabi encode params -v bool[] '[1,0,"false]' 
0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000
```

```bash
$ ethabi encode params -v bool[] '[1,"0,false]' 
000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000001
```
I looked at the parsing process of the `tokenize_array` method. I think it should handle the single quoted data here and return invalid data error.

### DIFF
--- a/ethabi/src/token/mod.rs
+++ b/ethabi/src/token/mod.rs
@@ -78,6 +78,10 @@ pub trait Tokenizer {
 			}
 		}
 
+		if ignore {
+			return Err(ErrorKind::InvalidData.into());
+		}
+
 		Ok(result)
 	}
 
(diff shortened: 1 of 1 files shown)


driftluo:
@debris I added a test for this situation.

Issue #113, opened 2018-07-16
Experimental ABIEncoderV2 support?


petarjs:
Hey guys, do you support the experimental ABIEncoderV2?

That would allow passing arrays and structs as Solidity function parameters.

Thanks!

debris:
We do not, but PRs are welcomed 😉

tkmct:
Is there someone working on this issue?
If not, I'd like to give it a try.

dvdplm:
@tkmct I don’t think so, have at it! :)

dvdplm:
@tkmct any progress on this on your end?

vkgnosis:
Fixed on master.

PR #142, opened 2018-12-16
fix(encode): fix dynamic array encoding; #113

hobofan:
Fixes the encoding of nested dynamic arrays, as a step towards support for the experimental ABIEncoderV2 (#113).

I had to adjust some existing tests, but I am pretty sure that they were incorrect to begin with. I confirmed that with the help of `web3.js`, and `ethers.js`, which both handle ABIEncoderV2 correctly.


hobofan:
@debris Is there anything stopping this from being merged? 😇

PR #145, opened 2019-02-06
Handle multiple events with the same name in one contract

leodasvacas:
By indexing the events by signature instead of name. We should somehow do the same for functions since currently they are also indexed by name, but functions don't have a `signature()` method available, so I'm not sure what to do there. A possible approach would be to to not use a map at all and use linear search on a vector.

This is a breaking change and will require a major release.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -213,7 +213,7 @@ fn hash_signature(sig: &str) -> Hash {
     sponge.update(&data);
     sponge.finalize(&mut result);
 
-    // This was deprecated but the replacement seems to not be availible.
+    // This was deprecated but the replacement seems to not be available.
     #[allow(deprecated)]
     Hash::from_slice(&result)
 }
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -17,6 +17,7 @@ use ethabi::param_type::{ParamType, Reader};
 use ethabi::token::{Token, Tokenizer, StrictTokenizer, LenientTokenizer};
 use ethabi::{encode, decode, Contract, Function, Event, Hash};
 use error::{Error, ResultExt};
+use std::str::FromStr;
 
 pub const ETHABI: &str = r#"
 Ethereum ABI coder.
@@ -27,7 +28,7 @@ Usage:
     ethabi encode params [-v <type> <param>]... [-l | --lenient]
     ethabi decode function <abi-path> <function-name> <data>
     ethabi decode params [-t <type>]... <data>
-    ethabi decode log <abi-path> <event-name> [-l <topic>]... <data>
+    ethabi decode log <abi-path> <event-signature> [-l <topic>]... <data>
     ethabi -h | --help
 
 Options:
@@ -51,7 +52,7 @@ struct Args {
 	cmd_log: bool,
 	arg_abi_path: String,
 	arg_function_name: String,
-	arg_event_name: String,
+	arg_event_signature: String,
 	arg_param: Vec<String>,
 	arg_type: Vec<String>,
 	arg_data: String,
@@ -89,7 +90,7 @@ fn execute<S, I>(command: I) -> Result<String, Error> where I: IntoIterator<Item
 	} else if args.cmd_decode && args.cmd_params {
 		decode_params(&args.arg_type, &args.arg_data)
 	} else if args.cmd_decode && args.cmd_log {
-		decode_log(&args.arg_abi_path, &args.arg_event_name, &args.arg_topic, &args.arg_data)
+		decode_log(&args.arg_abi_path, &Hash::from_str(&args.arg_event_signature)?, &args.arg_topic, &args.arg_data)
 	} else {
 		unreachable!()
 	}
@@ -102,10 +103,10 @@ fn load_function(path: &str, function: &str) -> Result<Function, Error> {
 	Ok(function)
 }
 
-fn load_event(path: &str, event: &str) -> Result<Event, Error> {
+fn load_event(path: &str, signature: &Hash) -> Result<Event, Error> {
 	let file = File::open(path)?;
 	let contract = Contract::load(file)?;
-	let event = contract.event(event)?.clone();
+	let event = contract.event(signature)?.clone();
 	Ok(event)
 }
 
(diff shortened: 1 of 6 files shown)


leodasvacas:
[clabot:check]

cli/src/main.rs
```diff
@@ -203,6 +205,19 @@ fn decode_log(path: &str, event: &str, topics: &[String], data: &str) -> Result<
...
 
+
+fn hash_signature(sig: &str) -> Hash {
+    let mut result = [0u8; 32];
+    let data = sig.replace(" ", "").into_bytes();
+    let mut sponge = Keccak::new_keccak256();
+    sponge.update(&data);
+    sponge.finalize(&mut result);
+
+    // This was deprecated but the replacement seems to not be availible.
```
Jannis: Typo: availible

leodasvacas: Thanks!

cli/src/main.rs
```diff
@@ -27,7 +29,7 @@ Usage:
     ethabi encode params [-v <type> <param>]... [-l | --lenient]
     ethabi decode function <abi-path> <function-name> <data>
     ethabi decode params [-t <type>]... <data>
-    ethabi decode log <abi-path> <event-name> [-l <topic>]... <data>
+    ethabi decode log <abi-path> <event-signature> [-l <topic>]... <data>
```
tomusdrw: IMHO this makes it much less human-friendly, only to handle an edge case of having two events with the same signature.

I'd go for supporting both name and signature here (i.e. assume name if `(` is not detected) - that should be possible with additional change I proposed below.

ethabi/src/contract.rs
```diff
@@ -14,8 +15,8 @@ pub struct Contract {
 	pub constructor: Option<Constructor>,
 	/// Contract functions.
 	pub functions: HashMap<String, Function>,
-	/// Contract events.
-	pub events: HashMap<String, Event>,
+	/// Contract events, maps signature to event.
+	pub events: HashMap<Hash, Event>,
```
tomusdrw: Can't we just store `HashMap<String, Vec<Event>>`? And maybe maintain additional `signature -> Event` mapping if you feel like looking inside the vec is too costly.

cli/src/main.rs
```diff
@@ -203,6 +205,19 @@ fn decode_log(path: &str, event: &str, topics: &[String], data: &str) -> Result<
 	Ok(result)
 }
 
+
+fn hash_signature(sig: &str) -> Hash {
```
tomusdrw: Should we do some pre-processing to make sure the signature is valid (beyond removing spaces)

ethabi/src/contract.rs
```diff
@@ -81,8 +82,9 @@ impl Contract {
 	}
 
 	/// Creates event decoder.
-	pub fn event(&self, name: &str) -> errors::Result<&Event> {
-		self.events.get(name).ok_or_else(|| ErrorKind::InvalidName(name.to_owned()).into())
+	pub fn event(&self, signature: &Hash) -> errors::Result<&Event> {
```
tomusdrw: Would add name-based search as well (returning a Vec)

tomusdrw:
CC @debris could say what you think about it?

ethabi/src/contract.rs
leodasvacas: That's a reasonable option.

cli/src/main.rs
leodasvacas: What do you have in mind?

cli/src/main.rs
```diff
@@ -284,7 +299,7 @@ bool false";
 
 	#[test]
 	fn log_decode() {
-		let command = "ethabi decode log ../res/event.abi Event -l 0000000000000000000000000000000000000000000000000000000000000001 0000000000000000000000004444444444444444444444444444444444444444".split(" ");
+		let command = "ethabi decode log ../res/event.abi Event(bool,address) -l 0000000000000000000000000000000000000000000000000000000000000001 0000000000000000000000004444444444444444444444444444444444444444".split(" ");
```
debris: `Event` is not ambiguous if there is only on event type with such name. Could you add support for it for backwards compatibility?

leodasvacas:
@tomusdrw @debris I've updated this to be backwards compatible with passing just a name, and to index the name to a vector of events, as per your suggestions. I'd like to do the same for functions but I'm not sure how that would work for the cli.

Jannis:
@tomusdrw @debris Any update on this one?

cli/src/main.rs
tomusdrw: I meant matching braces, invalid characters, etc. Since now by-signature is optional I don't care about it that match though.

ethabi/src/contract.rs
```diff
@@ -80,9 +81,10 @@ impl Contract {
 		self.functions.get(name).ok_or_else(|| ErrorKind::InvalidName(name.to_owned()).into())
 	}
 
-	/// Creates event decoder.
-	pub fn event(&self, name: &str) -> errors::Result<&Event> {
-		self.events.get(name).ok_or_else(|| ErrorKind::InvalidName(name.to_owned()).into())
+	/// Get all contract events named `name`.
+	pub fn event(&self, name: &str) -> errors::Result<&Vec<Event>> {
```
tomusdrw: I'd leave that method API unchanged and add a `events_by_name` function to return the vector.

leodasvacas:
@tomusdrw Thanks for taking the time to take another look at this, I've addressed your review.

cli/src/main.rs
```diff
@@ -203,6 +225,19 @@ fn decode_log(path: &str, event: &str, topics: &[String], data: &str) -> Result<
...
+
+fn hash_signature(sig: &str) -> Hash {
+    let mut result = [0u8; 32];
+    let data = sig.replace(" ", "").into_bytes();
+    let mut sponge = Keccak::new_keccak256();
+    sponge.update(&data);
+    sponge.finalize(&mut result);
+
+    // This was deprecated but the replacement seems to not be available.
+    #[allow(deprecated)]
```
dvdplm: Hmm, are you sure about that? It is [still there](https://github.com/paritytech/parity-common/blob/681faff5a201c63da132d8874191e862503fe010/fixed-hash/src/hash.rs#L199) and not deprecated. I do not get a warning locally.

leodasvacas: Thanks, probably a dependency update changed this.

tomusdrw:
Thanks a lot!

PR #147, opened 2019-04-01
Fix abi encoding/decoding

AtkinsChang:
- Fix dynamic array decoding while #142 fixes only encoding part
- Fix fixed-length array encoding/decoding


AtkinsChang:
[clabot:check]

tomusdrw:
I don't know enough about the ABI spec to be able to review that completely (unless you point me to the details). The code looks nice & clean though.

ethabi/src/decoder.rs
```diff
@@ -189,13 +189,29 @@ fn decode_param(param: &ParamType, slices: &[[u8; 32]], offset: usize) -> Result
...
-			let mut tokens = vec![];
-			let mut new_offset = offset;
-			for _ in 0..len {
-				let res = try!(decode_param(t, &slices, new_offset));
-				new_offset = res.new_offset;
-				tokens.push(res.token);
-			}
+			let mut tokens = Vec::with_capacity(len);
+			let new_offset = if param.is_dynamic() {
+				let offset_slice = try!(peek(slices, offset));
```
dvdplm: minor nitpick: new code should prefer `?` rather than `try!`

ethabi/src/token/token.rs
```diff
@@ -73,6 +73,15 @@ impl fmt::Display for Token {
 }
 
 impl Token {
+	/// returns wheater this token is dynamic
```
dvdplm:
```suggestion
	/// Returns whether this token is dynamic. `FixedArray` is considered dynamic if any of its members is dynamic.
```

Issue #146, opened 2019-03-08
Decoding Indexed Arrays (including strings and bytes) are not being decoded properly


davekaj:
As per the [Solidity Event spec](https://solidity.readthedocs.io/en/v0.5.5/contracts.html#events), it states the following about the EVM:

_If you use arrays (including `string` and `bytes`) as indexed arguments, its Keccak-256 hash is stored as a topic instead, this is because a topic can only hold a single word (32 bytes)._

I ran a test of this on Ropsten to confirm for myself, which shows 5 different ways indexed arrays are hashed: https://ropsten.etherscan.io/address/0xbc49d14d6f564bf3f9672aa1ba64f0166a29cde7#events

Right now, as it appears to me the decoder doesn't account for this. I have my own contract with an indexed string, and when trying to decode it it failed at this line: https://github.com/paritytech/ethabi/blob/9bf7261fd21192d9be56754b4cd75ecb9ff07bfb/ethabi/src/decoder.rs#L47

If we look at `event.rs`, we see where it might be failing:
https://github.com/paritytech/ethabi/blob/9bf7261fd21192d9be56754b4cd75ecb9ff07bfb/ethabi/src/event.rs#L112-L135

Specifically, at line 135 we are passing in the event types to decode as normal. But when these event fields are indexed arrays, they should actually check for a `bytes32` as the topic.

Please let me know if I am on the right track here, as I am not too familiar with this repository and I may have missed something

PR #148, opened 2019-04-02
Fix decoding indexed string, array and bytes parameters

Jannis:
Fixes https://github.com/paritytech/ethabi/issues/146.

The gist of #146 and this PR is: indexed string, array and bytes parameters are encoded and hashed using keccak256 and end up in the topic as `bytes32`. Because of this, we cannot decode them the usual way — we need to decode them using `ParamType::FixedBytes(32)`.

Without this, decoding any `Log` with indexed string, array or bytes parameters fails.

Note: Structs also fall in this category, but since they are not supported yet, this PR doesn't include anything for them.


Jannis:
[clabot:check]

ethabi/src/event.rs
```diff
@@ -124,7 +138,7 @@ impl Event {
 		};
 
 		let topic_types = topic_params.iter()
-			.map(|p| p.kind.clone())
+			.map(|p| self.convert_topic_param_type(&p.kind))
```
dvdplm: Am I reading this correctly: at this point we throw away the type information for `String`, `Bytes`, `Array` and `FixedArray` and they are all coerced into `FixedBytes`. I can see how that fixes decoding but isn't that valuable info? Like, how is the consuming code to know which type each param was? Or isn't that relevant in this context?

tomusdrw: @dvdplm it's not relevant, since we store hash of the data, not the data itself, hence we change the original type to `FixedBytes`

PR #150, opened 2019-04-09
Publish script.

tomusdrw:


tomusdrw:
@dvdplm yeah that could be nice. Even better if tagging a release would be the only thing we need to do and publishing is done by the CI. Currently I tag manually but will add it here.

dvdplm:
@tomusdrw tried this today and it didn't quite work. The error came from `derive`: it couldn't find `ethabi` version 8. Checking shows that `v8` was indeed published so I suspect there's a delay between publication and update of the registry so `derive` didn't "see" the new `ethabi` as published. Perhaps we can just pause 5-10secs after each?

Also, re-running it doesn't work because execution halts at `ethabi`: "already published"...

tomusdrw:
@dvdplm your machine is just too fast :P Indeed we can add a delay and maybe `cargo clean` at start.

dvdplm:
lol

tomusdrw:
see https://github.com/paritytech/ethabi/pull/153/files

PR #153, opened 2019-06-03
Improve publish script

tomusdrw:
1. Add 5 s delay before every crate and some logs
2. `cargo clean` at start


dvdplm:
It's still not idem-potent, but will probably fix the issue. We'll see! :)

Btw should we mention this in the README? Maybe not.

PR #160, opened 2019-10-30
added `payable` property from the ABI for functions and constructor

nlordell:
We could also do this for the fallback function as well, but that would be a breaking change as `Contract::fallback` would no longer return a bool but a `Fallback` struct (or something similar).


nlordell:
[clabot:check]

vkgnosis:
Closing because this has been fixed by 4c2b68bd41227b17ae6d79c6f2ee7cc7a50c81b0 .

PR #163, opened 2020-01-05
improved: better return code when failing

### LINKED ISSUE #162: Exit status of ethabi does not return error (non-zero) when an error is reported
When ethabi reports an error, such as invalid arguments, the return value to the system should be non-zero. The return value appears to always be 0, on both success and failure.

Here is a partial example of what I mean:

```
$ ethabi
Invalid arguments.

Usage:
    ethabi encode function <abi-path> <function-name> [-p <param>]... [-l | --lenient]
    ethabi encode params [-v <type> <param>]... [-l | --lenient]
    ethabi decode function <abi-path> <function-name> <data>
    ethabi decode params [-t <type>]... <data>
    ethabi decode log <abi-path> <event-name-or-signature> [-l <topic>]... <data>
    ethabi -h | --help

Caused by:
  Invalid arguments.

$ echo $?
0


$ ethabi encode params -v uint 1 -l
0000000000000000000000000000000000000000000000000000000000000001

$ echo $?
0


$ ethabi encode params -v uint 1
Cannot parse uint256

Caused by:
  Invalid input length

$ echo $?
0
```

cecton:
Fixes #162

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -66,7 +66,10 @@ fn main() {
 
 	match result {
 		Ok(s) => println!("{}", s),
-		Err(error) => print_err(&error),
+		Err(error) => {
+			print_err(&error);
+			std::process::exit(1);
+		},
 	}
 }
 


tomusdrw:
@ordian probably no one really though of that. I guess now we could do a bunch of improvements to the overall code as well since:
1. `print_err` is not required, there is `eprintln!` now afair
2. `fn main` can now return `Result` so we could easily just pass the return value and the error code would be handled automagically.

However it's not super high priority, might be a good intro task if someone wants to contribute to the codebase more.

cecton:
I guess for someone who wants to learn more about programming CLI with Rust but this doesn't seem like it will teach me much about ethabi or anything I don't know already (I might be wrong)

tomusdrw:
@cecton indeed, that's why there isn't even an issue for that. And I didn't really mean you when I was writing it, just left it here for posterity :)

dvdplm:
@cecton do you mind adding issues for Tomeks grumbles and perhaps label them "good first issue"?

PR #164, opened 2020-01-07
Replace docopt by structopt

### LINKED ISSUE #161: encoding many arguments runs out of memory!
```sh
export V=[{"constant":false,"gas":null,"inputs":[{"type":"address","name":""},{"type":"address","name":""},{"type":"uint256","name":""},{"type":"uint256","name":""},{"type":"uint256","name":""},{"type":"uint256","name":""},{"type":"address","name":""},{"type":"uint8","name":""},{"type":"bytes32","name":""},{"type":"bytes32","name":""},{"type":"address","name":""},{"type":"address","name":""},{"type":"uint256","name":""},{"type":"uint256","name":""},{"type":"bool","name":""},{"type":"uint8","name":""},{"type":"bytes32","name":""},{"type":"bytes32","name":""}],"name":"lotsofstuff","outputs":[],"payable":false,"stateMutability":null,"type":"function"}]
ethabi encode function <(echo $V) lotsofstuff -p 2C33B04ae6B405dEE94a6503dbe6a7CaD3f082Ca -p 2C33B04ae6B405dEE94a6503dbe6a7CaD3f082Ca -p 0000000000000000000000000000000000000000000000001bc16d674ec80000 -p 00000000000000000000000000000000000000000000000001671455cef6f770 -p 0000000000000000000000000000000000000000000000000000000000000000 -p 0000000000000000000000000000000000000000000000000000000000000000 -p c3455912Cf4bF115835A655c70bCeFC9cF4568eB -p 28 -p 7b93f2490c1833e65f9bfa58cd50e9f097edc408e65ebe88a5551dfb53ec3755 -p 4ab85d434db4881066e773411c6cf6b86966f4b5860f475b4dba2fa5608302ce -p 2C33B04ae6B405dEE94a6503dbe6a7CaD3f082Ca -p 2ef865ca017912c872826e25736a7a11f94235af -p 0000000000000000000000000000000000000000000000000000000000000000 -p 0000000000000000000000000000000000000000000000000000000000000000 -p 1 -p 28 -p 020303e12cbea2da5ab221a728bac53b8618cf8d9f3f4968cd24a7d9d1cf6abe -p 1e5a551d56422020a5b8f9b70d19b3771b0fc4c79334e353cef42a8eeeecc879 -l
```

For me running such a process on mac it is forcibly killed after ~2 minutes after eating up too much memory. (I thought rust was supposed to not do this :( ) what's going on here?

cecton:
Closes #161

I do this because docopt has quite big performance and memory issue and the maintainer recommends that we switch to clap. Hopefully this will fix the issue #161 (it does).


cli/src/main.rs
```diff
@@ -1,109 +1,109 @@
...
 use ethabi::token::{Token, Tokenizer, StrictTokenizer, LenientTokenizer};
 use ethabi::{encode, decode, Contract, Function, Event, Hash};
-use error::{Error, ErrorKind, ResultExt};
+use itertools::Itertools;
+use crate::error::{Error, ErrorKind, ResultExt};
 use tiny_keccak::Keccak;
 
-pub const ETHABI: &str = r#"
-Ethereum ABI coder.
-  Copyright 2016-2018 Parity Technologies (UK) Limited
```
dvdplm: I think we want the copyright to show still.

cli/src/main.rs
```diff
@@ -1,109 +1,109 @@
...
+#[derive(StructOpt, Debug)]
+enum Encode {
+	/// Load function from JSON ABI file.
+	Function {
+		abi_path: String,
+		function_name: String,
+		#[structopt(short, number_of_values = 1)]
+		params: Vec<String>,
+		#[structopt(short, long)]
+		lenient: bool,
```
tomusdrw: Some docs on `lenient` would be good.

cli/src/main.rs
```diff
@@ -1,109 +1,109 @@
-extern crate docopt;
 extern crate rustc_hex as hex;
```
tomusdrw: I guess we can get rid of all `extern crate`?

cli/src/main.rs
cecton: Done. And I changed it to 2019,

cli/src/main.rs
cecton: done

cli/src/main.rs
cecton: Done. But it renamed the hex import to `rustc_hex`.

cecton:
> One little annoyance is that the help output is now "nested" such that `ethabi -h` shows the subcomands but to see e.g. what options are valid for encoding a function I have to `ethabi encode function -h`. Is it possible to tweak that somehow without too much hassle?

I think it is possible to replace the entire usage message by a string so we could potentially use the old one. But if we do that, it won't update automatically if we add props or change some things. I'm not in favor.

cli/src/main.rs
```diff
@@ -1,109 +1,111 @@
...
-	arg_abi_path: String,
-	arg_function_name: String,
-	arg_event_name_or_signature: String,
-	arg_param: Vec<String>,
-	arg_type: Vec<String>,
-	arg_data: String,
-	arg_topic: Vec<String>,
-	flag_lenient: bool,
+#[derive(StructOpt, Debug)]
+#[structopt(about = "Copyright 2016-2019 Parity Technologies (UK) Limited")]
```
dvdplm: I think you can use `///` here for the about (for consistency with how the docs are specified below)?

cecton: Done. I also re-added the description of the CLI: Ethereum ABI coder.

PR #166, opened 2020-01-08
Add support for overloaded contract functions

Jannis:
Prior to this, ethabi would store just on function definition per function name in the `Contract` struct. This is not sufficient, in fact it is incorrect, as functions can be overloaded and there can be multiple definitions with the same name.

This applies the same treatment to contract functions that was applied earlier for events: for every name, a vector of function definitions is now stored. ethabi-cli takes either a name or a signature of the form

    functionName()
    functionName():(bool)
    functionName(uint256,string):(uint256,bool)

and matches this to the right overloaded version of the function with that name.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -29,7 +29,7 @@ enum Encode {
 	/// Load function from JSON ABI file.
 	Function {
 		abi_path: String,
-		function_name: String,
+		function_name_or_signature: String,
 		#[structopt(short, number_of_values = 1)]
 		params: Vec<String>,
 		/// Allow short representation of input params.
@@ -54,7 +54,7 @@ enum Decode {
 	/// Load function from JSON ABI file.
 	Function {
 		abi_path: String,
-		function_name: String,
+		function_name_or_signature: String,
 		data: String,
 	},
 	/// Specify types of input params inline.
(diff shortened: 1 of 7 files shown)


dvdplm:
Oh, one more thing: do you think you can add an example to the README as well? Thanks!

ethabi/src/contract.rs
```diff
@@ -112,7 +125,7 @@ impl Contract {
 }
 
 /// Contract functions interator.
```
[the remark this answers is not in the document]
Jannis: Done

Jannis:
@dvdplm All updated, thanks for the quick review!

cli/src/main.rs
```diff
@@ -87,25 +87,44 @@ where
...
+			contract.functions_by_name(name)?.iter().find(|f| {
+				f.signature() == name_or_signature
+			}).cloned().ok_or(ErrorKind::InvalidFunctionSignature(name_or_signature.to_owned()).into())
+		},
+
+		// It's a name
+		None => {
+			let functions = contract.functions_by_name(name_or_signature)?;
+			match functions.len() {
+				0 => unreachable!(),
```
tomusdrw: Can you elaborate why is that unreachable? What if I pass incorrect function name? I think previous code was just returning an error, wasn't it?

Jannis: I merely copied the logic that existed for events as well. Shall I make it return an error for both, functions and events, if no match exists?

tomusdrw: Yes please, if you don't mind, unless we can prove it doesn't panic if you give it a wrong name. Perhaps add a test case?

Jannis: Ah, it's `unreachable!` because `contract.functions_by_name(name_or_sig)?` returns an error if no function definition with that name exists. Same with the events. And they way these are built when parsing an ABI guarantees that no name exists with an empty list of events/functions.

```
➜  ethabi git:(jannis/overloaded-functions-1) cargo run -p ethabi-cli encode function res/test.abi 'baz' -p 1
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/ethabi encode function res/test.abi baz -p 1`
Error: Error(Ethabi(InvalidName("baz")), State { next_error: None, backtrace: InternalBacktrace })
```

Jannis: I've added two tests for running the CLI with a non-existent function name and event name.

tomusdrw: Awesome, thank you very much for looking into it!

PR #181, opened 2020-03-16
make input/output_param_types public

WyseNynja:
This is helpful for one of my projects.

### DIFF
--- a/ethabi/src/function.rs
+++ b/ethabi/src/function.rs
@@ -30,12 +30,12 @@ pub struct Function {
 
 impl Function {
 	/// Returns all input params of given function.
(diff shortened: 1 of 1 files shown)


dvdplm:
@WyseNynja thank you for the PR! Can you elaborate a bit more on why you want this change? Making fields public is generally frowned upon and we need to better understand your use-case before accepting.

WyseNynja:
I was decoding separately from where my function object was. I ended up refactoring to not need this.

WyseNynja:
Also, this just made functions that return cloned objects public. I don't think it exposes any actual internals.

PR #182, opened 2020-03-16
add TryInto impls

WyseNynja:


dvdplm:
Please add a description and an argument for why you want this.

WyseNynja:
Sure 😄 

`function.decode_output` returns a `Vec<ethabi::Token>`. These helpers make it easy to do something like this:

```
let x: U256 = returned_tokens[0].try_into()?
```

Digging into the code more. It looks like I could use these helpers instead do:
```
let x: U256 = returned_tokens[0].to_int()?
```

They seem to have the same goals.

PR #183, opened 2020-03-30
Update ethereum-types to 0.9

WyseNynja:
I am trying to use 0.9, but ethabi being on 0.8 blocks that.

I'm not sure about the version bump, but it seems like the right thing to do.


vorot93:
We export `ethereum-types` as a part of our API hence we need to bump to `12.0`.

dvdplm:
> Do you think you can bump `tiny-keccak` to v1.5 too? otherwise lgtm.

nvm I don't think it's needed right now.

Issue #178, opened 2020-01-23
Nested Tuple kind incorrect parsing


brockelmore:
When doing a `type_check` on this tuple:
```
struct Order {
        address makerAddress;           // Address that created the order.
        address takerAddress;           // Address that is allowed to fill the order. If set to 0, any address is allowed to fill the order.
        address feeRecipientAddress;    // Address that will recieve fees when order is filled.
        address senderAddress;          // Address that is allowed to call Exchange contract methods that affect this order. If set to 0, any address is allowed to call these methods.
        uint256 makerAssetAmount;       // Amount of makerAsset being offered by maker. Must be greater than 0.
        uint256 takerAssetAmount;       // Amount of takerAsset being bid on by maker. Must be greater than 0.
        uint256 makerFee;               // Fee paid to feeRecipient by maker when order is filled.
        uint256 takerFee;               // Fee paid to feeRecipient by taker when order is filled.
        uint256 expirationTimeSeconds;  // Timestamp in seconds at which order expires.
        uint256 salt;                   // Arbitrary number to facilitate uniqueness of the order's hash.
        bytes makerAssetData;           // Encoded data that can be decoded by a specified proxy contract when transferring makerAsset. The leading bytes4 references the id of the asset proxy.
        bytes takerAssetData;           // Encoded data that can be decoded by a specified proxy contract when transferring takerAsset. The leading bytes4 references the id of the asset proxy.
        bytes makerFeeAssetData;        // Encoded data that can be decoded by a specified proxy contract when transferring makerFeeAsset. The leading bytes4 references the id of the asset proxy.
        bytes takerFeeAssetData;        // Encoded data that can be decoded by a specified proxy contract when transferring takerFeeAsset. The leading bytes4 references the id of the asset proxy.
    }

struct APIOrder {
        Order[] orders;
        uint256 makerAssetFillAmount;
        bytes[] signatures;
        address exchange;
        uint256 protocolFee;
}

struct APIOrders {
        APIOrder[6] api_orders;
  }
```

input types are returned as: `Tuple([FixedArray(Tuple([]), 6)])` instead of the full type. There is also an issue around decoded deeply nested tuples and dynamic length bytes but I will open a separate issue for that

brockelmore:
Seems like this may be a fix for [here](https://github.com/paritytech/ethabi/blob/a985fb9f8c38bc7953f83500c92a0eda62b52f08/ethabi/src/tuple_param.rs#L77):
```
[...]

PR #186, opened 2020-04-03
Add support for nested tuple arrays and fixed arrays

nlordell:
The `ethabi` crate currently does not support tuples with nested tuple arrays and fixed arrays of the form
```solidity
struct Bar {
    address a;
}
struct Foo {
    Bar[] bar;
    Bar[2] bar_fixed;
}
```

This PR adds support for it. There is a bit of duplicated logic between `TupleParam` and `Param` deserialization code, but I didn't address that here.

### Test Plan

Added a new unit test for testing nested tuple types.

### DIFF
--- a/ethabi/src/param.rs
+++ b/ethabi/src/param.rs
@@ -235,4 +235,49 @@ mod tests {
 			}
 		);
 	}
+
+	#[test]
+	fn param_tuple_with_nested_tuple_arrays_deserialization() {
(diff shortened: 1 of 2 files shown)


nlordell:
Partially fixes #178

vkgnosis:
Included through 12e704bafc5ec53a48901edacda9294a34ebcbfe

PR #187, opened 2020-04-08
Support the receive function as an additional operation type

fordN:
In the Solidity 0.6.0 release the unnamed function, previously `fallback`, has been split into two functions, `receive` and `fallback`.   ([0.6.0 breaking changes](https://solidity.readthedocs.io/en/v0.6.0/060-breaking-changes.html))

This PR adds support for the `receive` function.

### DIFF
--- a/ethabi/src/contract.rs
+++ b/ethabi/src/contract.rs
@@ -117,6 +117,11 @@ impl Contract {
 	pub fn events(&self) -> Events {
 		Events(self.events.values().flatten())
 	}
+
+	/// Returns true if contract has fallback
+	pub fn fallback(&self) -> bool {
+		self.fallback
+	}
 }
 
 /// Contract functions iterator.
--- a/ethabi/src/contract.rs
+++ b/ethabi/src/contract.rs
@@ -17,7 +17,7 @@ use std::iter::Flatten;
 use std::{fmt, io};
 
 /// API building calls to contracts ABI.
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, Default, PartialEq)]
 pub struct Contract {
 	/// Contract constructor.
 	pub constructor: Option<Constructor>,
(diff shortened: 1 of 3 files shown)


ethabi/src/contract.rs
```diff
@@ -47,12 +49,14 @@ impl<'a> Visitor<'a> for ContractVisitor {
...
-		A: SeqAccess<'a>,
-	{
-		let mut result =
-			Contract { constructor: None, functions: HashMap::default(), events: HashMap::default(), fallback: false };
+	fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error> where A: SeqAccess<'a> {
+		let mut result = Contract {
+			constructor: None,
+			functions: HashMap::default(),
+			events: HashMap::default(),
+			receive: true,
```
gakonst, line 57: Should the default `receive` value be true?

dvdplm: I have the same question. Reading [this](https://solidity.readthedocs.io/en/v0.6.0/060-breaking-changes.html#semantic-and-syntactic-changes) I'd say it should be `false`.
Somewhat related q: why not impl `Default` for `Contract`?

ethabi/src/contract.rs
```diff
@@ -116,11 +123,6 @@ impl Contract {
 	pub fn events(&self) -> Events {
 		Events(self.events.values().flatten())
 	}
-
-	/// Returns true if contract has fallback
-	pub fn fallback(&self) -> bool {
```
dvdplm, line 121: Why remove this?

fordN: It's not used anywhere.

ethabi/src/contract.rs
fordN: Hmm, good point, I'm honestly not sure why I thought `true` was the right default here.  It should indeed be `false`. 

As for impl `Default`, with this PR I was trying to be as unintrusive as possible to get it in quickly, but we are way past that point ;).  Looking at it now: the default values for `Contract` properties are consistent with the defaults for each property type so `Default` can be derived.

gakonst:
@dvdplm could you take another look at this? Changes seems straightforward to me

ethabi/src/operation.rs
```diff
@@ -25,6 +25,7 @@ pub enum Operation {
 	Event(Event),
 	/// Fallback, ignored.
 	Fallback,
+	Receive,
```
dvdplm, line 28: Please document.

tomusdrw:
@fordN could you please add a test case and docs @dvdplm requested?

smessmer:
is there any progress on this? This is blocking https://github.com/tomusdrw/rust-web3/issues/410 which in turn blocks one of my projects.

vkgnosis:
Merged in 90659eb0b260ea079ca7a5e52a8ca04c2e2a810d

PR #193, opened 2020-11-05
Change the versions in Cargo.toml to indicate this is a fork

lutter:
Add `-graph` to all versions so that there is no ambiguity between this fork and upstream. This seems to be the key to getting around a very annoying [issue](https://github.com/rust-analyzer/rust-analyzer/issues/6407) with `rust-analyzer` where it complains about missing imports from this fork.


lutter:
This was completely the wrong repo. Ignore

PR #194, opened 2020-11-20
Deserialize array of tuples from contract ABI.

dvush:
This allows to deserialize contract ABI with arrays of structs.

example:
```solidity
pragma solidity ^0.7.0;
pragma experimental ABIEncoderV2;

contract Example {
    struct Struct {
        uint param1;
        bytes param2;
    }

    function foo(Struct[] memory _s) external {
        // ... 
    }

    function bar(Struct[10] memory _s) external {
        // ... 
    }
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ target
 *.swp
 *.swo
 .vscode
+.idea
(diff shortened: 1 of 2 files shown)


vkgnosis:
Has been fixed by one of the commits in https://github.com/rust-ethereum/ethabi/pull/212 . The test case works.

PR #197, opened 2020-12-04
Fix clippy warnings

vorot93:


nlordell (approved):
Looks good. My only comment is that since the method names on the `Token` type changed, this will require a major version bump.

PR #203, opened 2021-02-02
avoid OOM on decoding corrupted array encodings

f0rki:
~~Currently ethabi can crash when parsing corrupted input data containing an array. This is due to an out-of-memory error when allocating a huge vector upfront. This can be easily avoided by checking whether the array length is possible for the given input before allocating the vector.~~

Not relevant anymore, but maybe the testcase I wrote is?


f0rki:
my original PR doesn't seem relevant anymore. The testcase I added is now passing with the changes from v14. Maybe you'd like to add this testcase?

vkgnosis (approved):
The test case looks useful to me, thanks.

nlordell (approved):

f0rki:
I pushed another fix (and a testcase) to this branch. There was a bounds check missing during decoding of dynamic tuples.

nlordell (approved):
Excellent!

PR #210, opened 2021-03-13
Change Token::Int associated data from Uint to Int

rhinodavid:
Both `Uint` and `Int` are aliased to `ethereum_types::U256`, but semantically it seems like the associated data for `Token::Int` should be an `Int`.


vkgnosis (approved):

nlordell:
Note that
```rust
pub type Int = ethereum_types::U256;
```

nlordell (approved):

PR #211, opened 2021-03-15
Support `no_std`

AurevoirXavier:
Hi! Would you like to make this crate no-std?


VargSupercolony:
any updates on that? we really want to use it in ink! contracts :(

AurevoirXavier:
> any updates on that? we really want to use it in ink! contracts :(

Should I update the codebase?

But seems no one review this.

sorpaas:
Please rebase and I'll do a review.

PR #215, opened 2021-03-26
Update dependencies

vkgnosis:
Based on output of `cargo outdated --root-deps-only`.

### DIFF
--- a/ethabi/Cargo.toml
+++ b/ethabi/Cargo.toml
@@ -18,7 +18,7 @@ hex = "0.4"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 sha3 = "0.9"
-ethereum-types = "0.10.0"
+ethereum-types = "0.11.0"
 thiserror = "1"
 uint = "0.9.0"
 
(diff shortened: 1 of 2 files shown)


ethabi/Cargo.toml
```diff
@@ -18,7 +18,7 @@ hex = "0.4"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 sha3 = "0.9"
-ethereum-types = "0.10.0"
+ethereum-types = "0.11.0"
```
nlordell, line 21: This is interface breaking in that crates that use `ethereum_types::U256` on with version `0.10.0` won't work with this crate.

May be worth considering re-exporting these types here so upstream crates can use the re-exported versions and avoid needing to keep multiple version in-sync?

nlordell (approved):
:tada:

ethabi/Cargo.toml
vkgnosis: Agree we should consider re-exporting. Won't make it part of this PR but will make an issue for it.

PR #226, opened 2021-06-25
Serialize contracts

taminomara:
Implement serialization for `contract::Contract` struct.

This is required for [gnosis/ethcontract-rs/170](https://github.com/gnosis/ethcontract-rs/issues/170).

### DIFF
--- a/ethabi/src/contract.rs
+++ b/ethabi/src/contract.rs
@@ -7,6 +7,7 @@
 // except according to those terms.
 
 use crate::{errors, operation::Operation, Constructor, Error, Event, Function};
+use serde::ser::SerializeSeq;
 use serde::{
 	de::{SeqAccess, Visitor},
 	Deserialize, Deserializer, Serialize, Serializer,
@@ -16,7 +17,6 @@ use std::{
 	fmt, io,
 	iter::Flatten,
 };
-use serde::ser::SerializeSeq;
 
 /// API building calls to contracts ABI.
 #[derive(Clone, Debug, Default, PartialEq)]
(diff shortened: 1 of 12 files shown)


taminomara:
Tests are still in progress.

ethabi/src/contract.rs
```diff
@@ -199,3 +199,373 @@ impl<'a> Iterator for Events<'a> {
 		self.0.next()
 	}
 }
+
+#[cfg(test)]
+mod test {
```
nlordell, line 204: Thanks for all the tests!

ethabi/src/param.rs
```diff
@@ -213,24 +249,23 @@ mod tests {
...
 
 	#[test]
-	fn param_tuple_array_deserialization() {
+	fn param_tuple_array() {
 		let s = r#"{
 			"name": "foo",
 			"type": "tuple[]",
 			"components": [
 				{
-					"name": "amount",
```
nlordell, line 164: Maybe I missed something, but I didn't see a change related to this - was it just the test was failing on main?

taminomara: Yeah, we don't save these names in rust struct, so when we serialize it back to json, they're lost. I've removed them so that `assert_json_eq` could pass.

taminomara:
The clippy check fails on code that wasn't modified in this PR. I'll fix this in a separate PR and rebase this one.

taminomara:
Fixed clippy in #227.

vkgnosis (approved):
Looking at the existing deserialization code there are probably a bunch of types that could make use of deriving it instead of manually implementing. Not related to this PR which LGTM, just a note.

taminomara:
Branch rebased, all checks passed.

taminomara:
> a bunch of types that could make use of deriving it instead of manually implementing

I've tried to use deriving where possible. Are there anything I've missed?

vkgnosis:
I mean the existing manual Deserialize implementations. I didn't look too hard but it felt like some of them could make use of https://serde.rs/attributes.html to derive in more places. Don't think we have to look at it for this PR though.

nlordell (approved):

PR #231, opened 2021-08-11
Add optional field internalType to Param

### LINKED ISSUE #230: Add support for the internalType (Solidity 0.5.11+) of the abi
Solidity compiler at version [0.5.11](https://github.com/ethereum/solidity/blob/develop/Changelog.md#0511-2019-08-12) added a `internalType` field in the abi inputs/outputs. 

This field contains the type as declared in the solidity code, this is useful to determine the actually struct type of a tuple.
for exmaple:

```json
{
  "components": [
    {
      "components": [..]
      "internalType": "struct Pairing.G1Point",
      "name": "A",
      "type": "tuple"
    }
  ],
  "internalType": "struct Verifier.Proof",
  "name": "proof",
  "type": "tuple"
}
```

mattsse:
Add `internalType` field to `Param`.

- Closes #230

### DIFF
--- a/derive/src/function.rs
+++ b/derive/src/function.rs
@@ -407,7 +407,7 @@ mod tests {
 						outputs: vec![ethabi::Param {
 							name: "".to_owned(),
 							kind: ethabi::ParamType::Uint(256usize),
-							internal_type :None
+							internal_type: None
 						}, ethabi::Param {
 							name: "".to_owned(),
 							kind: ethabi::ParamType::String,
--- a/derive/src/function.rs
+++ b/derive/src/function.rs
@@ -289,8 +289,12 @@ mod tests {
 		#[allow(deprecated)]
 		let ethabi_function = ethabi::Function {
 			name: "hello".into(),
-			inputs: vec![ethabi::Param { name: "foo".into(), kind: ethabi::ParamType::Address, internal_type :None }],
-			outputs: vec![ethabi::Param { name: "bar".into(), kind: ethabi::ParamType::Uint(256), internal_type :None }],
+			inputs: vec![ethabi::Param { name: "foo".into(), kind: ethabi::ParamType::Address, internal_type: None }],
+			outputs: vec![ethabi::Param {
+				name: "bar".into(),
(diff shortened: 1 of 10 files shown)


gakonst (approved):
LGTM - @vkgnosis for context, we need this for ABI Encoder v2 struct generation

ethabi/src/param.rs
```diff
@@ -212,6 +246,42 @@ mod tests {
...
+			"internalType": "struct Pairing.G1Point[]",
+			"components": [
+				{
+					"type": "uint48"
+				},
+				{
+					"type": "tuple",
+					"components": [
+						{
+							"type": "address"
```
nlordell, line 270: Can tuple components have internal types as well? It would make sense in the case of something like:
```solidity
struct Foo {
  IERC20 token;
  Bar bar;
}
```
For both `bar` and `token` fields to also have an `internalType`.

mattsse: yes, this field is added for every single component regardless of type, struct, primitive, enum.
and since a component contains a `components: Vec<Component>`,  this goes as long as how deeply nested the struct is.

nlordell (approved):

vkgnosis (approved):

gakonst:
Is this OK to merge? @mattsse

derive/src/function.rs
```diff
@@ -389,17 +397,21 @@ mod tests {
...
 							name: "bar".to_owned(),
-							kind: ethabi::ParamType::Array(Box::new(ethabi::ParamType::Uint(256usize)))
+							kind: ethabi::ParamType::Array(Box::new(ethabi::ParamType::Uint(256usize))),
+							internal_type: None
 						}],
 						outputs: vec![ethabi::Param {
 							name: "".to_owned(),
-							kind: ethabi::ParamType::Uint(256usize)
+							kind: ethabi::ParamType::Uint(256usize),
+							internal_type :None
```
sorpaas, line 410:
```suggestion
							internal_type: None
```

PR #236, opened 2021-09-08
Non-Standardized Supporting

AurevoirXavier:
I revamp my old implementation #211.

- All JSON tests were ignored under no-std. 
- `Reader`/`Tokenizer`/`serde`/`serde_json`should be enabled by `--features full-serde`.
  - I'm not sure if I should disable `Writer` in no-std, because I found it was used in `signature`.
- `Error::FromUtf8Error` removed, cause can't find any usage of it.
- All `HashMap` change to `BTreeMap` for no-std. **!!BREAKING change!!**

@sorpaas

### DIFF
--- a/ethabi/Cargo.toml
+++ b/ethabi/Cargo.toml
@@ -27,14 +27,21 @@ hex-literal = "0.3"
 paste = "1"
 
 [features]
-default = ["std"]
+default = [
+	"std",
+	"full-serde"
+]
 std = [
 	"anyhow/std",
 	"hex/std",
-	"serde",
-	"serde_json",
 	"sha3/std",
 	"ethereum-types/std",
 	"thiserror",
 	"uint/std",
+]
+
+# To enable custom `Reader` and `serde` features support
+full-serde = [
+	"serde",
+	"serde_json",
 ]
--- a/ethabi/Cargo.toml
+++ b/ethabi/Cargo.toml
@@ -13,14 +13,14 @@ description = "Easy to use conversion of ethereum contract calls to bytecode."
 edition = "2018"
 
 [dependencies]
-anyhow = { version = "1", default-features = false }
+anyhow = { version = "1", optional = true }
 hex = { version = "0.4", default-features = false, features = ["alloc"]}
 serde = { version = "1.0", optional =  true, features = ["derive"] }
 serde_json = { version = "1.0", optional = true }
 sha3 = { version = "0.9", default-features = false }
 ethereum-types = { version = "0.11.0", default-features = false }
 thiserror = { version = "1", optional = true }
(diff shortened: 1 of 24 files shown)


ethabi/src/contract.rs
```diff
@@ -6,33 +6,45 @@
...
 // except according to those terms.
 
-use crate::{errors, operation::Operation, Constructor, Error, Event, Function};
+#[cfg(not(feature = "std"))]
+use alloc::collections::{btree_map::Values, BTreeMap};
+#[cfg(feature = "full-serde")]
+use core::fmt;
+use core::iter::Flatten;
+#[cfg(feature = "std")]
+use std::collections::{btree_map::Values, BTreeMap};
```
sorpaas, line 15: You technically don't need this line.

Add `extern crate alloc;` and the above `alloc` import will work on `std` as well.

sorpaas:
I'd appreciate if you can add CI build run for no_std as well!

vkgnosis:
Could you explain a bit on the motivation for no std support? What part of the crate is still useful without std? Why does HashMap have to change to BTreeMap?

AurevoirXavier:
> Could you explain a bit on the motivation for no std support? What part of the crate is still useful without std? Why does HashMap have to change to BTreeMap?

I found many people use this in substrate runtime, so do I.

Some of them used it in EVM. And my team used it for the ethereum cross-chain bridge.

AurevoirXavier:
> I'd appreciate if you can add CI build run for no_std as well!

Okay.

sorpaas:
Please remove check and test command in CI, and add build.

AurevoirXavier:
> Please remove check and test command in CI, and add build.

Please review.

sorpaas:
CI is failing!

AurevoirXavier:
> CI is failing!

I see. `cargo-hack` is pretty strict.

ethabi/Cargo.toml
```diff
@@ -28,7 +28,6 @@ paste = "1"
 
 [features]
 default = [
-	"std",
```
sorpaas: This is just wrong. By default `std` should be enabled.

AurevoirXavier: > This is just wrong. By default `std` should be enabled.

But the `full-serde` contains std. Should I explicitly write it here again?

AurevoirXavier: Okay. I found a corner case. You're right.

AurevoirXavier: Sorry about ^^

The `full-serde` is enabled by `default`. So `std` is always enabled by `defualt`.

sorpaas (approved):

PR #238, opened 2021-09-11
Remove `anyhow` in crates intended to be libraries

c410-f3r:
`anyhow` is mostly used for applications and the usage of such crate in a library brings unnecessary weight and additional compilation times.

> This library provides anyhow::Error, a trait object based error type for easy idiomatic error handling in Rust **APPLICATIONS**.

### DIFF
--- a/ethabi/Cargo.toml
+++ b/ethabi/Cargo.toml
@@ -13,7 +13,6 @@ description = "Easy to use conversion of ethereum contract calls to bytecode."
 edition = "2018"
 
 [dependencies]
-anyhow = { version = "1", optional = true }
 hex = { version = "0.4", default-features = false, features = ["alloc"] }
 serde = { version = "1.0", optional =  true, features = ["derive"] }
 serde_json = { version = "1.0", optional = true }
@@ -33,7 +32,6 @@ default = [
 	"rlp",
 ]
 std = [
-	"anyhow/std",
 	"hex/std",
 	"sha3/std",
 	"ethereum-types/std",
@@ -44,7 +42,6 @@ std = [
 # To enable custom `Reader`/`Tokenizer` and `serde` features support
 full-serde = [
 	"std",
-	"anyhow",
 	"serde",
 	"serde_json",
 	"uint",
--- a/ethabi/src/errors.rs
+++ b/ethabi/src/errors.rs
@@ -6,16 +6,13 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-#[cfg(feature = "full-serde")]
-use core::num;
-
-#[cfg(feature = "full-serde")]
-use anyhow::anyhow;
-#[cfg(feature = "std")]
-use thiserror::Error;
-
+use crate::no_std_prelude::Cow;
 #[cfg(not(feature = "std"))]
 use crate::no_std_prelude::*;
+#[cfg(feature = "full-serde")]
+use core::num;
+#[cfg(feature = "std")]
+use thiserror::Error;
 
 /// Ethabi result type
 pub type Result<T> = core::result::Result<T, Error>;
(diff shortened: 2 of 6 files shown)


ethabi/src/errors.rs
```diff
@@ -34,17 +33,20 @@ pub enum Error {
...
 	#[error("Hex parsing error: {0}")]
 	Hex(#[from] hex::FromHexError),
-	/// Other errors.
+	/// Other errors (Static string).
 	#[error("{0}")]
-	Other(#[from] anyhow::Error),
+	OtherStr(&'static str),
+	/// Other errors (Dynamic string).
+	#[error("{0}")]
+	OtherString(String),
```
vkgnosis, line 41: I would like there to only be one Other type. Maybe that could be Cow<'static, str> ?.

vkgnosis (commented):
Anyhow removal makes sense to me. For the default features (which mostly removes the `std` features) I'd wait until the other PR that enabled no-std support is merged which does the same thing.

ethabi/src/errors.rs
c410-f3r: Done

c410-f3r:
Which PR? There are two `no_std` related PRs but they are closed. In any case, the changes here are only intented to take off some dependency burden and should be seen as an occasional and complementary step towards `no_std`

ethabi/Cargo.toml
```diff
@@ -13,14 +13,13 @@ description = "Easy to use conversion of ethereum contract calls to bytecode."
...
-sha3 = { version = "0.9", default-features = false }
-ethereum-types = { version = "0.12.0", default-features = false }
+hex = { default-features = false, version = "0.4" }
+serde = { default-features = false, version = "1.0", features = ["derive"], optional = true }
+serde_json = { default-features = false, features = ["std"], version = "1.0", optional = true }
+sha3 = { default-features = false, version = "0.9" }
+ethereum-types = { default-features = false, version = "0.12" }
 thiserror = { version = "1", optional = true }
-uint = { version = "0.9.0", optional = true }
+uint = { default-features = false, optional = true, version = "0.9" }
```
vkgnosis, line 22: Could you keep the ordering of version and default-features as before so it is clear in the diff that for example `sha3` hasn't changed?

Also while I agree about the anyhow removal I am hesitant to remove other features because doing this broke something for users. See https://github.com/rust-ethereum/ethabi/pull/240  and https://github.com/rust-ethereum/ethabi/pull/239 .
So I feel we would first have to check if those features have any impact on users even this crate still compiles. Or better yet add tests to make sure the features we want to provide are there like ethreum-types/serde on full-serde.

c410-f3r: As requested, the order was kept, features weren't modified and only anyhow was removed

vkgnosis (approved):
LGTM. Will wait for more opinions on the PR before merging.

ethabi/src/lib.rs
```diff
@@ -17,12 +17,16 @@ extern crate alloc;
...
 		string::{self, String, ToString},
 		vec::Vec,
 	};
 }
+#[cfg(feature = "std")]
+mod no_std_prelude {
+	pub use std::borrow::Cow;
+}
 #[cfg(not(feature = "std"))]
 use no_std_prelude::*;
```
c410-f3r, line 31: It is easier to simply `extern crate alloc` regardless of regardless of `no_std` and then  call `core ...`, `alloc ...` and `std ...` where appropriated instead of writing custom modules imports

c410-f3r:
Rebased. Any news about this PR?

vkgnosis:
I like it but I'd also like to get a second opinion.

c410-f3r:
> I like it but I'd also like to get a second opinion.

Thanks

nlordell (approved):
Seems like a good change.

> It is easier to simply extern crate alloc regardless of no_std and then call core ..., alloc ... and std ... where appropriated instead of writing custom modules imports

I also agree. I do think its a bit out of the scope of this PR, but definitely a nice future improvement.

PR #244, opened 2021-09-17
Add a method to retrieve the short 4 byte signature of a function

### LINKED ISSUE #243: Function should have a method to retrieve the 4 byte signature
Function has been written towards encoding rust types for sending transactions.  But decoding on-chain bytes given an ABI is another important use case.  

One blocker is Function cannot tell you what the 4 byte signature is.  Doing it 'yourself' is hard because the signature module is private (so one would have to re-invent that code), and the 'input_param_types()' method is also private.  

Hashing the results of 'signature' isn't sufficient since ETH smart contracts don't use the return type in the hash.

ryanobjc:
Fixes #243

### DIFF
--- a/ethabi/src/function.rs
+++ b/ethabi/src/function.rs
@@ -63,6 +63,12 @@ impl Function {
 		Ok(signed.into_iter().chain(encoded.into_iter()).collect())
 	}
 
+	/// Return the 4 byte short signature of this function.
+	pub fn short_signature(&self) -> [u8; 4] {
+		let params = self.input_param_types();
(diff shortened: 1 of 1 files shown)


ryanobjc:
I don't care what we call the function name, I just went with something I hoped was reasonable.

vkgnosis (approved):

ryanobjc:
The pull requests says "1 workflow awaiting approval - first time contributors need a maintainer to approve running workflows."

nlordell (approved):

PR #251, opened 2021-12-06
feat: add support for solidity error type

### LINKED ISSUE #225: Support error type
There doesn't seem to exist support for the new `error` type that is defined in the latest ABI spec.

Consider the followign solidity code:

```solidity
pragma solidity ^0.8.5;

error DummyError(uint256 reason);

contract Foo {

  function bar() public pure {
    revert DummyError({reason: 4});
  }
}
```

The ABI that `solc` generates from this is:

```json
[
   {
      "inputs":[
         {
            "internalType":"uint256",
            "name":"reason",
            "type":"uint256"
         }
      ],
      "name":"DummyError",
      "type":"error"
   },
   {
      "inputs":[
         
      ],
      "name":"bar",
      "outputs":[
         
      ],
      "stateMutability":"pure",
      "type":"function"
   }
]
```

Feeding that to `ethabi` results in the following error:

>"Invalid operation type error."

So, I guess `error` is just not yet supported as a `type`.

mattsse:
Add support for solidity error type, closes #225

following this spec

https://docs.soliditylang.org/en/v0.8.10/abi-spec.html#errors

codec is the same as functions

Added serde tests and exposed similar functions for error on the `Contract`

### DIFF
--- a/derive/src/contract.rs
+++ b/derive/src/contract.rs
@@ -75,6 +75,7 @@ mod test {
 			constructor: None,
 			functions: Default::default(),
 			events: Default::default(),
+			errors: Default::default(),
 			receive: false,
 			fallback: false,
 		};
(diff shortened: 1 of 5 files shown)


gakonst:
@vkgnosis @nlordell any chance we can get some eyes on this?

nlordell (approved):
Changes look good to me.

PR #255, opened 2022-01-08
add ether units parsing to lenient uint tokenizer

joshieDo:
#### Motivation

Allowing `LenientTokenizer` to parse `uint256` with units: `ether|gwei|nano|nanoether|wei`

eg: [Foundry](https://github.com/gakonst/foundry) uses the `LenientTokenizer` to parse function arguments given by the user through its CLI. This would help out when inserting large values. 

I don't know if inserting units into this library is desirable... The reason I'm proposing the change here is for cases such as arrays : `[1ether, 1gwei]`, which would be quite troublesome to parse in higher level libraries.

### DIFF
--- a/ethabi/src/token/lenient.rs
+++ b/ethabi/src/token/lenient.rs
@@ -78,7 +78,7 @@ impl Tokenizer for LenientTokenizer {
 							return Err(dec_error.into());
 						}
 
-						Uint::from_dec_str(&float_n.round().to_string())
+						Uint::from_dec_str(&float_n.to_string())
 							.map_err(|_| Error::Other(Cow::Owned(original_dec_error)))?
 					}
 					None => return Err(dec_error.into()),
--- a/ethabi/src/token/lenient.rs
+++ b/ethabi/src/token/lenient.rs
@@ -14,7 +14,8 @@ use crate::{
 use std::borrow::Cow;
 
 use once_cell::sync::Lazy;
-static RE: Lazy<regex::Regex> = Lazy::new(|| regex::Regex::new(r"([0-9]+(?:\.[0-9]+)?)\s*(ether|gwei|nanoether|nano|wei)").expect("invalid regex"));
+static RE: Lazy<regex::Regex> =
+	Lazy::new(|| regex::Regex::new(r"([0-9]+(?:\.[0-9]+)?)\s*(ether|gwei|nanoether|nano|wei)").expect("invalid regex"));
 
 /// Tries to parse string as a token. Does not require string to clearly represent the value.
 pub struct LenientTokenizer;
--- a/ethabi/src/token/lenient.rs
+++ b/ethabi/src/token/lenient.rs
@@ -11,8 +11,14 @@ use crate::{
 	token::{StrictTokenizer, Tokenizer},
 	Uint,
 };
-use regex::Regex;
 use std::borrow::Cow;
+
+macro_rules! regex {
+	($re:literal $(,)?) => {{
+		static RE: once_cell::sync::OnceCell<regex::Regex> = once_cell::sync::OnceCell::new();
+		RE.get_or_init(|| regex::Regex::new($re).unwrap())
+	}};
+}
 
 /// Tries to parse string as a token. Does not require string to clearly represent the value.
 pub struct LenientTokenizer;
@@ -51,8 +57,8 @@ impl Tokenizer for LenientTokenizer {
 			Ok(_uint) => _uint,
 			Err(dec_error) => {
 				let original_dec_error = dec_error.to_string();
-				let re = Regex::new(r"([0-9.]+) *(ether|gwei|nano|nanoether|wei)")
-					.map_err(|_| Error::Other(Cow::Borrowed("invalid regex expression")))?;
+				let re = regex!(r##"([0-9]+\.{0,1}[0-9]+)\s*(ether|gwei|nano|nanoether|wei)"##);
+
 				match re.captures(value) {
 					Some(captures) => {
 						if captures.len() != 3 {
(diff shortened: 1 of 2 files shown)


gakonst:
I'm also supportive of this feature, it'd help us a lot. Main point I made to @joshieDo was that we could consider implementing this with a manual parser (given it's simple enough) instead of adding an extra dependency

ethabi/src/token/lenient.rs
```diff
@@ -11,9 +11,15 @@ use crate::{
...
-use regex::Regex;
 use std::borrow::Cow;
 
+macro_rules! regex {
+	($re:literal $(,)?) => {{
+		static RE: once_cell::sync::OnceCell<regex::Regex> = once_cell::sync::OnceCell::new();
+		RE.get_or_init(|| regex::Regex::new($re).unwrap())
+	}};
+}
+
```
gakonst, line 22: Think you could also just do:

```suggestion
use once_cell::sync::Lazy;

static RE: Lazy<regex::Regex> = Lazy::new(|| regex::Regex::new(r"([0-9]+\.{0,1}[0-9]+)\s*(ether|gwei|nano|nanoether|wei)").expect("invalid regex"));
```


and just do `RE.matches`

gakonst: instead of doing the macro

gakonst:
Hi @nlordell think we can get CI to run here / a look at the PR?

ethabi/src/token/lenient.rs
```diff
@@ -43,7 +47,49 @@ impl Tokenizer for LenientTokenizer {
...
+		let uint = match Uint::from_dec_str(value) {
+			Ok(_uint) => _uint,
+			Err(dec_error) => {
+				let original_dec_error = dec_error.to_string();
+
+				match RE.captures(value) {
+					Some(captures) => {
+						if captures.len() != 3 {
+							return Err(dec_error.into());
+						} else {
```
vkgnosis, line 62: Is it possible that the length is not 3? If we have this number of groups in the regex it would make sense that the regex cannot have matched if the number of groups is wrong. Likewise below we have `.get` and return an error if we cannot get the group but shouldn't that be unwrap because it cannot happen?

ethabi/src/token/lenient.rs
```diff
@@ -13,6 +13,11 @@ use crate::{
 };
 use std::borrow::Cow;
 
+use once_cell::sync::Lazy;
+static RE: Lazy<regex::Regex> = Lazy::new(|| {
+	regex::Regex::new(r"^([0-9]+(?:\.[0-9]+)?)\s*(ether|gwei|nanoether|nano|wei)$").expect("invalid regex")
```
vkgnosis, line 18: What is the purpose of `?:\.`? Shouldn't it be just `\.`?

vkgnosis (commented):
Code looks fine. I'm not sure I can judge whether it makes sense for this component or not though.

ethabi/src/token/lenient.rs
joshieDo: `(?:` makes it a non-capturing group. Otherwise, `2.2ether` would have the following groups: `2`,`.2`,`ether`

vkgnosis (approved):

gakonst:
Nice! Think we can merge? @nlordell @vkgnosis @joshieDo

ethabi/src/token/lenient.rs
```diff
@@ -74,3 +110,119 @@ impl Tokenizer for LenientTokenizer {
...
+		);
+	}
+
+	#[test]
+	fn tokenize_uint_wei() {
+		assert_eq!(LenientTokenizer::tokenize(&ParamType::Uint(256), "1wei").unwrap(), Token::Uint(Uint::from(1)));
+
+		assert_eq!(LenientTokenizer::tokenize(&ParamType::Uint(256), "1 wei").unwrap(), Token::Uint(Uint::from(1)));
+
+		assert_eq!(LenientTokenizer::tokenize(&ParamType::Uint(256), "0.1 wei").unwrap(), Token::Uint(Uint::from(0)));
```
nlordell, line 142: Shouldn't this error too? Feels a bit unexpected to me.

nlordell:
Looks good to me! Just have one small discussion point then I think we can merge.

ethabi/src/token/lenient.rs
joshieDo: Added something to handle beyond 18 decimals. However, I do get some precision issues:

```
ether_string | ether_to_f64_to_ether | f64.fract()
"0.0000000000000000001" | "0.0000000000000000001" | 0.09999999999999999
"1.0000000000000000001" | "1" | 0.0
```

Added a string comparison, which is not the prettiest, so I'm open to suggestions.

nlordell: Hmm, good point. I think you can't rely on `f64` here because you will run into precision issues with ETH.

What I would recommend is to parse the whole and fractional parts as `U256`s and then add them together. In pseudo-code:
```rust
let whole = captures.get(1).expect("capture group does not exist").as_str();
// requires additional capture group for the fractional part
let frac = captures.get(2).map(|c| c.as_str().trim_start('.')).unwrap_or_else("");
let units = captures.get(3).expect("capture group does not exist").as_str();

let whole_pow = match units.to_lowercase().as_str() {
    "ether" => 18u32,
    "gwei" | "nano" | "nanoether" => 9,
    "wei" => 0,
    _ => return Err(dec_error.into()),
};
let frac_pow = frac.len().checked_sub(whole_pow as _).ok_or_else(|| dec_error)?;

whole.checked_mul(U256::from(10).pow(whole_pow))
    .and_then(|whole| {
        whole.checked_add(frac.checked_mul(U256::from(10).pow(frac_pow))?)
     })
    .ok_or_else(|| dec_error)
```

joshieDo: You're right, it feels way better like this. Thanks!

nlordell (approved):
LGTM.

PR #256, opened 2022-01-11
fix: `function.constant` as an `Option<bool>`

brockelmore:
As noted in the deprecation warning, when parsing compiler output, ethabi fills in a `false`. This causes trouble in `ethers` and `foundry` by making it so compiler outputted ABIs aren't parseable by `ethers.js` due to bad `constant` settings (i.e. `false` for a `pure`).

This PR fixes that by keeping it concrete for derive, but making it optional in the main implementation. This allows us to parse json from the solc compiler correctly and pipe that out to a valid ABI json. This could be an incorrect implementation, but it seems to work

### DIFF
--- a/derive/src/function.rs
+++ b/derive/src/function.rs
@@ -61,8 +61,7 @@ pub struct Function {
 	/// Function output params.
 	outputs: Outputs,
 	#[deprecated(note = "The constant attribute was removed in Solidity 0.5.0 and has been \
-				replaced with stateMutability. If parsing a JSON AST created with \
-				this version or later this value will always be false, which may be wrong.")]
+				replaced with stateMutability.")]
 	/// Constant function.
 	constant: bool,
 	/// Whether the function reads or modifies blockchain state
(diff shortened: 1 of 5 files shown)


brockelmore:
an alternative method to fix this would be to peak at the `state_mutability` and make sure they are compliant with each other but an `Option` feels cleaner

gakonst (approved):

nlordell (approved):
LGTM

vkgnosis (approved):

gakonst:
good to merge

filiplazovic:
https://github.com/gakonst/ethers-rs/blob/master/ethers-core/src/abi/human_readable.rs#L320 <-- this is now not compiling since it pulls `ethabi` from Github

gakonst:
The `ethabi` is locked to a specific commit, so shouldn't be happening? In either case, we should fix there, PRs / Issue to track is welcome!

PR #257, opened 2022-01-12
fix(reader): default to u8 for unmatched types

### LINKED ISSUE #254: Ethabi cannot parse library with public function
# Problem

It seems that ethabi cannot parse the ABI of a library contract with a `public` function (i.e. one that's intended to be linked)

This was originally reported in https://github.com/gakonst/foundry/issues/349.

# Repro

```solidity
pragma solidity >=0.7.6;

library LibTest {
    enum TestEnum {
        A,
        B,
        C
    }

    function foo(TestEnum test) public view {}
}
```

`solc contract.sol --abi` generates:

```json
[{"inputs":[{"internalType":"enum LibTest.TestEnum","name":"test","type":"LibTest.TestEnum"}],"name":"foobar","outputs":[],"stateMutability":"view","type":"function"}]
```

And then doing

```rust
let abi = r#"[{"inputs":[{"internalType":"enum LibTest.TestEnum","name":"test","type":"LibTest.TestEnum"}],"name":"foobar","outputs":[],"stateMutability":"view","type":"function"}]
"#;
let abi: ethabi::Contract = serde_json::From_str(abi).unwrap();
```

panics with:

```
panicked at 'called `Result::unwrap()` on an `Err` value: Error("InvalidName(\"LibTest.TestEnum\")", line: 1, column: 167)
```

## Diagnosis

The error originates in [this](https://github.com/rust-ethereum/ethabi/blob/master/ethabi/src/param_type/reader.rs#L160-L162) match branch. Custom type names are seemingly not supported.

```rust
let result = match name {
			"address" => ParamType::Address,
			"bytes" => ParamType::Bytes,
			"bool" => ParamType::Bool,
			"string" => ParamType::String,
			"int" => ParamType::Int(256),
			"tuple" => ParamType::Tuple(vec![]),
			"uint" => ParamType::Uint(256),
			s if s.starts_with("int") => {
				let len = s[3..].parse()?;
				ParamType::Int(len)
			}
			s if s.starts_with("uint") => {
				let len = s[4..].parse()?;
				ParamType::Uint(len)
			}
			s if s.starts_with("bytes") => {
				let len = s[5..].parse()?;
				ParamType::FixedBytes(len)
			}
			_ => {
				return Err(Error::InvalidName(name.to_owned()));
			}
```
# Solution

Can we assume that custom names that are not tuples are enums? If so, it could make sense to do `_ => ParamType::Uint(8)`. I'm not sure that assumption is correct. It'd be nice if we could access the `internalType` inside `Reader`, since it contains the word `enum`.

gakonst:
closes #254 by defaulting to uint8 on the remaining cases, which as discussed in the issue should always correspond to a Solidity enum.

### DIFF
--- a/ethabi/src/param_type/reader.rs
+++ b/ethabi/src/param_type/reader.rs
@@ -159,7 +159,9 @@ impl Reader {
 			}
 			// As discussed in https://github.com/rust-ethereum/ethabi/issues/254,
 			// any type that does not fit the above corresponds to a Solidity
-			// enum, and as a result we treat it as a uint8.
+			// `enum`, and as a result we treat it as a `uint8`. This is a unique
+			// case which occurs in libraries with exported (`public`) Solidity
+			// functions with `enum` parameters.
 			_ => ParamType::Uint(8),
 		};
 
--- a/ethabi/src/param_type/reader.rs
+++ b/ethabi/src/param_type/reader.rs
@@ -157,9 +157,10 @@ impl Reader {
 				let len = s[5..].parse()?;
 				ParamType::FixedBytes(len)
 			}
-			_ => {
-				return Err(Error::InvalidName(name.to_owned()));
-			}
+			// As discussed in https://github.com/rust-ethereum/ethabi/issues/254,
+			// any type that does not fit the above corresponds to a Solidity
+			// enum, and as a result we treat it as a uint8.
+			_ => ParamType::Uint(8),
 		};
 
 		Ok(result)


ethabi/src/param_type/reader.rs
```diff
@@ -157,9 +157,10 @@ impl Reader {
 				let len = s[5..].parse()?;
 				ParamType::FixedBytes(len)
 			}
-			_ => {
-				return Err(Error::InvalidName(name.to_owned()));
-			}
+			// As discussed in https://github.com/rust-ethereum/ethabi/issues/254,
+			// any type that does not fit the above corresponds to a Solidity
+			// enum, and as a result we treat it as a uint8.
```
nlordell, line 162: Maybe include in the comment that this is very specific to libraries with exported functions with enum parameters.

nlordell (approved):

ethabi/src/param_type/reader.rs
gakonst: done https://github.com/rust-ethereum/ethabi/pull/257/commits/95387acb79d7c969b1e07c5ac092cc2d8d681549

gakonst:
should be good to merge

ethabi/src/param_type/reader.rs
```diff
@@ -157,9 +157,12 @@ impl Reader {
...
 				ParamType::FixedBytes(len)
 			}
-			_ => {
-				return Err(Error::InvalidName(name.to_owned()));
-			}
+			// As discussed in https://github.com/rust-ethereum/ethabi/issues/254,
+			// any type that does not fit the above corresponds to a Solidity
+			// `enum`, and as a result we treat it as a `uint8`. This is a unique
+            // case which occurs in libraries with exported (`public`) Solidity
+            // functions with `enum` parameters.
```
nlordell, line 164:
```suggestion
		// case which occurs in libraries with exported (`public`) Solidity
		// functions with `enum` parameters.
```

nlordell:
> should be good to merge

Cool. I'll just give @vkgnosis a chance to look over it too.

ethabi/src/param_type/reader.rs
gakonst: oops

gakonst:
Doc formatting fixed.

@nlordell Sounds good! Will wait for @vkgnosis approval / merge.

vkgnosis (approved):
Code LGTM . I don't feel confident to judge whether change is right otherwise but I trust your discussion on it.

gakonst:
@mattsse WDYT? I think should address the issue / not introduce new bugs. I'd vote we optimistically merge and if somebody complains and turns out it's a hard to solve issue we revert.

mattsse:
yeah based on #254, this seems like the most appropriate fix

gakonst:
great! can we merge then? @nlordell

PR #259, opened 2022-01-29
Skip writing out of `function.constant` if `None`

brockelmore:
Skips serialization of `None` value for `constant` in function. Currently produces `null`, which can break some tooling that isn't expecting it, apparently 

Partial fix to https://github.com/gakonst/foundry/issues/618


gakonst (approved):

carlosdp:
Just ran into this, this makes abis generated by `forge` incompatible with The Graph deployments, which is strange since it's also using this library for parsing.

vkgnosis (approved):

gakonst:
Thanks @vkgnosis! Can we merge this?

PR #260, opened 2022-02-03
Add parity-codec cargo feature.

dandanlen:
This is useful if you want to use ethabi in a substrate-related project, for example abi-encoding eth contract calls on-chain.

### DIFF
--- a/ethabi/Cargo.toml
+++ b/ethabi/Cargo.toml
@@ -55,3 +55,7 @@ full-serde = [
 rlp = [
 	"ethereum-types/rlp",
 ]
+
+parity-codec = [
+	"ethereum-types/codec"
+]


vkgnosis:
Since nothing in this crate changes with the feature, wouldn't it make more sense that you depend on ethereum-types/codec when you need the feature there rather than going through this crate?

dandanlen:
I could, but then I need to add an explicit dependency on `ethereum-types`, which is currently only a transitive dependency via ethabi. 

For example, say I want to use `Address` type from ethabi. I don't really care that it is re-exported from `ethereum-types`, I just want it to be compatible with the ethabi crate.

I suppose I agree in that this feature would make more sense / would be more complete if it also conditionally derived the `Encode` and `Decode` traits for `Function`, `Param` etc.

vkgnosis (approved):
Makes sense to me. Will give the PR some time in case other people have an opinion.

gakonst (approved):
lgtm

nlordell (approved):
LGTM

gakonst:
If you merge this, it'd also help to release a new version of the crate!

PR #265, opened 2022-03-01
chore: loosen feature requirement on deser methods

gakonst:
the `full-serde` feature is not really necessary here, you can deserialize just with `serde` enabled and you don't gain anything by feature gating `reader` behind full-serde

### DIFF
--- a/ethabi/src/param_type/mod.rs
+++ b/ethabi/src/param_type/mod.rs
@@ -8,15 +8,13 @@
 
 //! Function and event param types.
 
-#[cfg(feature = "full-serde")]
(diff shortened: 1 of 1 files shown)


vkgnosis:
Should this be included in the release I'm making at the moment?

gakonst:
NVM this unfortunately doesn't seem to work. Both of these depend on `std` features, encountered while trying to trim dependencies from parity-common. The 2 PRs which should "properly" address the issue are: https://github.com/paritytech/parity-common/pull/630 and https://github.com/paritytech/parity-common/pull/625.

PR #266, opened 2022-03-13
Optimize encoder to minimize allocations and copying

ArtificialPB:
Original implementation had lots of unnecessary allocations and copying when encoding tokens to bytes. The updated code now pre-sizes the end vector of words to avoid resizing, uses references where possible, and appends encoded data directly into end vector, avoiding intermediary vector allocations.

The updated code is ~50% faster based on (naive) benchmark, and allocates much less memory:
```rust
use ethabi::token::{LenientTokenizer, Token, Tokenizer};
use std::time::{SystemTime, UNIX_EPOCH};

fn main() {
	let input = get_input();

	let loops = 5;
	let iterations = 100000;

	let mut total_time_ns = 0;
	let mut output = vec![];

	for _ in 0..loops {
		let start_nano = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_nanos();
		for _ in 0..iterations {
			output = ethabi::encode(&input);
		}
		let end_nano = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_nanos();
		total_time_ns += (end_nano - start_nano)
	}

	println!("Took: {} ns/op", total_time_ns / iterations / loops);
	println!("Total: {} ns", total_time_ns);
	println!("Output size: {} bytes", output.len());
}

fn get_input() -> Vec<Token> {
	let string1 = Token::String("test".to_owned());
	let string2 = Token::String("cyborg".to_owned());
	let string3 = Token::String("night".to_owned());
	let string4 = Token::String("day".to_owned());
	let string5 = Token::String("weee".to_owned());
	let string6 = Token::String("funtests".to_owned());
	let address =
		Token::Address(LenientTokenizer::tokenize_address("25f8087EAD173b73D6e8B84329989A8eEA16CF73").unwrap().into());
	let bool = Token::Bool(true);
	let fixed_array = Token::FixedArray(vec![Token::Uint(2331.into()), Token::Uint(93523141.into()), Token::Uint(538841.into())]);
	let dynamic_array = Token::Array(vec![Token::Uint(2331.into()), Token::Uint(93523141.into()), Token::Uint(538841.into())]);
	let deep_tuple = Token::Tuple(vec![string5, string6]);
	let inner_tuple = Token::Tuple(vec![string3, string4, fixed_array, deep_tuple]);
	let outer_tuple = Token::Tuple(vec![string1, bool, string2, inner_tuple, address, dynamic_array]);

	return vec![outer_tuple];
}
```

### DIFF
--- a/ethabi/src/encoder.rs
+++ b/ethabi/src/encoder.rs
@@ -12,15 +12,26 @@
 use crate::no_std_prelude::*;
 use crate::{util::pad_u32, Bytes, Token, Word};
 
-fn pad_bytes(bytes: &[u8]) -> Vec<Word> {
-	let mut result = vec![pad_u32(bytes.len() as u32)];
-	result.extend(pad_fixed_bytes(bytes));
-	result
+fn pad_bytes_len(bytes: &[u8]) -> u32 {
+	// "+ 1" because len is also appended
+	((bytes.len() + 31) / 32) as u32 + 1
 }
 
-fn pad_fixed_bytes(bytes: &[u8]) -> Vec<Word> {
+fn pad_bytes_append(data: &mut Vec<Word>, bytes: &[u8]) {
+	data.push(pad_u32(bytes.len() as u32));
+	fixed_bytes_append(data, bytes);
+}
+
+fn pad_fixed_bytes_len(bytes: &[u8]) -> u32 {
+	((bytes.len() + 31) / 32) as u32
+}
+
+fn pad_fixed_bytes_append(result: &mut Vec<Word>, bytes: &[u8]) {
+	fixed_bytes_append(result, bytes);
+}
+
+fn fixed_bytes_append(result: &mut Vec<Word>, bytes: &[u8]) {
 	let len = (bytes.len() + 31) / 32;
-	let mut result = Vec::with_capacity(len);
 	for i in 0..len {
 		let mut padded = [0u8; 32];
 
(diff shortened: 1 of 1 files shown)


vkgnosis:
Thanks for the PR. Just approved CI to run. Still reviewing it.

ArtificialPB:
One remaining issue I see with the new `RawArray` Mediate type, it currently assumes all its elements are of type `Raw`, which is invalid since it can also contain `RawTuple` type. This type should probably be handled the same way as `RawTuple`.

I'll add a test case to verify.

ArtificialPB:
Fixed the issue with `RawArray`. It's now handled the same way as `RawTuple`. 

Performance improvement remains unchanged.

vkgnosis (approved):
Code looks correct to me. I also think making the code a little more complicated to avoid many allocations makes sense. What was the motivation for this change? Did you specifically notice that this function was slow?

ArtificialPB:
@vkgnosis I'm optimizing our internal codebase, and noticed that the encoding via `ethabi` was a couple of times slower than the current JVM-based library we're using, which seemed a bit weird.

Are you open to further optimizations? Not just encoding, but potentially other parts of the code as well.

vkgnosis:
This code base has accumulated a lot of warts that make it slower, harder to use, harder to develop than it has to be. It is also maintenance mode where we mostly fix small issues as they come up and don't touch the code otherwise. The historical context is that it used to be actively developed by Parity (the company) who eventually gave it over to other volunteers.

I've had in the back of my mind for a while that it would be nice to start from scratch to do things properly without worrying about backward compatibility. At least for the primary functionality that the crate provides which is ethereum abi encoding/decoding and supplying the relevant basic types like U256. So if you want to invest a significant amount of time then that might be a more efficient way of going about it.

On the other I hand I'm sure everyone using this crate would appreciate further speed improvements to this crate and we are open to including them.

ArtificialPB:
Thanks for the info. At the moment we have no plans for a full rewrite, but we're happy to contribute to existing projects.

I'll keep opening additional PR's for optimizations that we might do on our fork.

gelfand (approved):

nlordell (approved):
LGTM

PR #269, opened 2022-06-09
feat: add Serialize trait to Token, Log and LogParams

### LINKED ISSUE #268: Add Serialize to Logs
This change is required in order to be able to Serialize events into json as per https://github.com/foundry-rs/foundry/issues/1432.

Genysys:
This PR closes #268

### DIFF
--- a/ethabi/src/log.rs
+++ b/ethabi/src/log.rs
@@ -6,6 +6,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+#[cfg(feature = "full-serde")]
 use serde::{Deserialize, Serialize};
 
 #[cfg(not(feature = "std"))]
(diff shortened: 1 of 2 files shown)


vkgnosis (commented):
Makes sense to me. Looking at other code that uses serde I believe this should be feature gated behind the full-serde feature.

Genysys:
Thanks for coming back quickly on this @vkgnosis . Can you please link an example of codebase that uses serde as such so I can amend the PR

nlordell:
For example: https://github.com/rust-ethereum/ethabi/blob/451d029c69a1fce7c4d57cc117e985a31dc53627/ethabi/src/event.rs#L25

vkgnosis (commented):
The `use serde::{Deserialize, Serialize};` lines also have to be feature gated.

vkgnosis (approved):

PR #277, opened 2022-08-12
Release v17.2.0

shekhirin:


cli/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ethabi-cli"
-version = "17.1.0"
+version = "17.2.0"
```
nlordell, line 3: Just curious, why not a minor version bump (since its just a bug fix for nested tuple and array parsing)?

shekhirin: thought that parsing nested arrays/tuples is a new feature rather than a bugfix, but can change to patch bump if you want to

nlordell: I'm not so familiar with the `ethabi` CLI tool, so I was just curious about the reasoning, not necessarily suggesting it should be one way or the other.

nlordell (approved):

vkgnosis (approved):

vkgnosis:
I'll make the crates.io release.

vkgnosis:
Turns out nlordell and I are only owners of the main ethabi crate and not the other crates (ethabi-cli, ethabi-contract, ethabi-derive) so we can't release those.

vkgnosis:
Since it's a binary crate maybe ethabi-cli doesn't need to be on crates.io anyway? You could use it from git.

PR #284, opened 2022-11-04
Decode function, which fails, if some data left

leviska:
Right now `decode` will return ok, even if some input data is left. 
I'm writing a decoder, which tries to decode the tx input and logs not knowing exact signature, so it tries several signatures and picks whatever decodes successfully. But current implementation is wrong in this context: it can return "Ok", even if the data is not fully corresponds to the types provided

I suggest adding some function, which will also check, that it decodes everything, or returns error, because implementing it requires implementation details (knowing offset). Also I need to decode logs, so I've added additional implementation there too, and if I missed some other decodes, I will happily add other functions too

Current implementation is just the easiest way I thought this could be done, but I'm open for review and will fix it if you have a better idea. Naming is for debate too

### DIFF
--- a/ethabi/src/decoder.rs
+++ b/ethabi/src/decoder.rs
@@ -39,9 +39,7 @@ fn as_bool(slice: &Word) -> Result<bool, Error> {
 	Ok(slice[31] == 1)
 }
 
-/// Decodes ABI compliant vector of bytes into vector of tokens described by types param.
-/// Returns ok, even if some data left to decode
-pub fn decode(types: &[ParamType], data: &[u8]) -> Result<Vec<Token>, Error> {
+fn decode_offset(types: &[ParamType], data: &[u8]) -> Result<(Vec<Token>, usize), Error> {
 	let is_empty_bytes_valid_encoding = types.iter().all(|t| t.is_empty_bytes_valid_encoding());
 	if !is_empty_bytes_valid_encoding && data.is_empty() {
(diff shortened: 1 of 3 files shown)


vkgnosis:
The change makes sense to me but we should probably do it in a backward compatible way. I'd keep the existing public functions as they are now and add new functions with the extra check.

leviska:
I've added some basic tests and left the original functions as is

vkgnosis:
One more nitpick. Sorry if I didn't make that clear in the previous comment. We should refactor out the common (copy pasted) parts between the base function and the _whole addition. What I meant before was that the behavior of the base function shouldn't change. We *can* change the implementation. Maybe the base function can call _whole internally or we need a third function that can be called by both.

leviska:
that's what I did in the first attempt, so I didn't quite get what you're were asking for 😅

isn't the first version what you suggesting? 
we make `decode_offset` (private function), and call it from `decode` and `decode_whole`, and similar with events.

I tried to use errors, but then I'll need to add another error in the enum and that's probably not the most backward compatible way(?)

image

vkgnosis:
You're right. I must have not paid attention when I read the first commit and thought it changed the behavior of the existing function. Just looked at the commit again and it looks fine.

leviska:
No problem :) 

I've returned the old version

vkgnosis (approved):

PR #288, opened 2022-11-17
decode verify, which checks that decoded data is equal to input

leviska:
Continue of https://github.com/rust-ethereum/ethabi/pull/284

The issue is, that for some types we still can check, that decoded data is not equal to encoded, but we need to check the actual data, not only offsets

There is an example of such types in test:

`FixedBytes(20)` will still be encoded as 32 bytes, but the data is in the first 20 bytes
`Address` will be encoded as 32 bytes, but the data is in the last 20 bytes

So to determine which type is the input data, we need to look at the data itself. And imo the easiest way to do that by encoding the data back and check, that everything is equal

### DIFF
--- a/ethabi/src/decoder.rs
+++ b/ethabi/src/decoder.rs
@@ -10,7 +10,7 @@
 
 #[cfg(not(feature = "std"))]
 use crate::no_std_prelude::*;
-use crate::{encode, Error, ParamType, Token, Word};
+use crate::{Error, ParamType, Token, Word};
 
 #[derive(Debug)]
 struct DecodeResult {
@@ -32,14 +32,11 @@ fn as_usize(slice: &Word) -> Result<usize, Error> {
 }
 
 fn as_bool(slice: &Word) -> Result<bool, Error> {
-	if !slice[..31].iter().all(|x| *x == 0) {
-		return Err(Error::InvalidData);
-	}
-
+	check_zeroes(&slice[..31])?;
 	Ok(slice[31] == 1)
 }
 
-fn decode_offset(types: &[ParamType], data: &[u8]) -> Result<(Vec<Token>, usize), Error> {
+fn decode_impl(types: &[ParamType], data: &[u8], validate: bool) -> Result<(Vec<Token>, usize), Error> {
 	let is_empty_bytes_valid_encoding = types.iter().all(|t| t.is_empty_bytes_valid_encoding());
 	if !is_empty_bytes_valid_encoding && data.is_empty() {
 		return Err(Error::InvalidName(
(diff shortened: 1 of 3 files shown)


vkgnosis:
Is there any standard that says the extra bytes *have* to be zero? It also seems possible that the contents of the extra bytes should be ignored in which case it shouldn't error.

leviska:
By solidity specifications it must be zeroes. 
https://docs.soliditylang.org/en/v0.8.17/abi-spec.html#formal-specification-of-the-encoding

image

vkgnosis:
This checks relies on our encoding doing the zeroing correctly so we can compare the results. The most correct way of doing this would be to check the zeroes in the decoding internals directly.

Doing it with the encoding round trip is slower and could be wrong if there is anything the spec leaves ambiguous. Because then an equality comparison wouldn't be enough to invalidate the other encoding.

Compared to your previous change, this check you could also make yourself outside this library because it doesn't rely on data that is private to the decode function. With the mentioned downsides I feel that is a better place. On the other hand if we did the validation directly in decode then I would be fine with it.  That's probably more involved and we'd have to decide how we want to handle backwards compatibility.

leviska:
> Compared to your previous change, this check you could also make yourself outside this library because it doesn't rely on data that is private to the decode function. 

For decode, yes
For logs, sadly, no(

leviska:
If there is a way to make it work logs, I'd take an easier way. If not, I could potentially implement something more complex

leviska:
I've implemented correct validation for some types.
Right now I left the `int` case, because I'm not sure what current implementation returns
In the doc it says, that in `int` case the value is two complemented and first 2 bytes are the sign bytes, but none of this is found in current implementation, so... Does the current implementation just returns the raw data, without actually parsing the integer? 
image
image

ethabi/src/decoder.rs
```diff
@@ -32,14 +32,11 @@ fn as_usize(slice: &Word) -> Result<usize, Error> {
 }
 
 fn as_bool(slice: &Word) -> Result<bool, Error> {
-	if !slice[..31].iter().all(|x| *x == 0) {
-		return Err(Error::InvalidData);
-	}
-
+	check_zeroes(&slice[..31])?;
```
vkgnosis, line 35: In other places we have `validate: bool` to toggle the behavior. Should we also do this here?

vkgnosis:
>  Does the current implementation just returns the raw data, without actually parsing the integer?

I don't know.

ethabi/src/decoder.rs
leviska: this check was already here, so the old `decode` was checking that too, so by adding a toggle we could change old behavior

leviska:
I think we can leave `int` case for now
Do you have any other suggestions?

vkgnosis (approved):

leviska:
Hi, can we merge this?

vkgnosis:
Is there a rush to merge it? We probably can but I also want to give other people a chance to review, even if it's unlikely that someone will.

PR #293, opened 2023-04-12
Allocate Vec capacity based on size hints during decoding

jordy25519:
Discovered this issue while profiling another crate

### DIFF
--- a/ethabi/src/filter.rs
+++ b/ethabi/src/filter.rs
@@ -50,9 +50,10 @@ impl Serialize for TopicFilter {
 }
 
 /// Acceptable topic possibilities.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, Default, PartialEq, Eq)]
 pub enum Topic<T> {
 	/// Match any.
+	#[default]
 	Any,
 	/// Match any of the hashes.
 	OneOf(Vec<T>),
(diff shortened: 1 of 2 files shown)


vkgnosis (approved):

vkgnosis (changes requested):
This fails CI. I guess there is a case where the code without with_capacity never end ups really allocating that much.

jordy25519:
> This fails CI. I guess there is a case where the code without with_capacity never end ups really allocating that much.

Ok I'm glad this happened since some malicious payloads could cause large memory allocation in my previous commit. Iterator should have the same effect to avoid reallocs and not allocate upfront

vkgnosis:
> Iterator should have the same effect to avoid reallocs and not allocate upfront

I don't think that's true. The current version of the PR works the same as it did before the PR but using `collect` instead of manually pushing. Either way the Vector reallocate as elements get added because the final size isn't known upfront.

ethabi/src/filter.rs
```diff
@@ -82,12 +83,6 @@ impl<T> Topic<T> {
 	}
 }
 
-impl<T> Default for Topic<T> {
-	fn default() -> Self {
-		Topic::Any
-	}
-}
```
vkgnosis, line 89: Why this change? I think you are changing behavior because the previous version still implemented Default when T wasn't Default while the new one doesn't. Either way, it's unrelated to the rest of the  PR.

jordy25519: I only changed it due to the CI failure, happy to keep or revert it either way

vkgnosis (approved):

PR #298, opened 2023-06-28
Fix #297: add name and internal type information to ParamType::Tuple

### LINKED ISSUE #297: Support field information in nested tuple field.
The current inputs and outputs params for `Function` and `Event` are `Vec<Param>`, but the problem is that, if the `Param` is a tuple, then will lose name and internal_type information about the elements in this tuple.

For example, if I have an abi like.
```json
   "inputs": [
      {
        "components": [
          {
            "internalType": "uint8",
            "name": "blockType",
            "type": "uint8"
          },
          {
            "components": [
              {
                "internalType": "uint256",
                "name": "txIndex",
                "type": "uint256"
              },
              {
                "internalType": "bytes",
                "name": "data",
                "type": "bytes"
              }
            ],
            "internalType": "struct ExchangeData.AuxiliaryData[]",
            "name": "auxiliaryData",
            "type": "tuple[]"
          },
          {
            "internalType": "bytes",
            "name": "offchainData",
            "type": "bytes"
          }
        ],
        "internalType": "struct ExchangeData.Block[]",
        "name": "blocks",
        "type": "tuple[]"
      }
    ],
```
This abi can be parsed fine, but the problem is that the current `ParamType` for tuple is `Tuple(Vec<ParamType>)`. So this means I cannot get the elements' names `blockType` and  `blocks`.

Those information is important in my use case, I would like to suggest to change the `ParamType::Tuple(Vec<ParamType>)` to `ParamType::Tuple(Vec<TupleParam>)`, in this way, users can get the name and internal type information.

al8n:
As mentioned in #297, this PR aims for it.

### DIFF
--- a/ethabi/src/decoder.rs
+++ b/ethabi/src/decoder.rs
@@ -240,7 +240,7 @@ fn decode_param(param: &ParamType, data: &[u8], offset: usize, validate: bool) -
 			let len = t.len();
 			let mut tokens = Vec::with_capacity(len);
 			for param in t {
-				let res = decode_param(param, tail, new_offset, validate)?;
+				let res = decode_param(&param.kind, tail, new_offset, validate)?;
 				new_offset = res.new_offset;
 				tokens.push(res.token);
 			}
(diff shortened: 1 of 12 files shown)


vkgnosis:
Don't ping individuals. If you need this urgently you can depend on a local fork of this project where you can make any changes you want to. This is a breaking change in the API so I don't know if we want to do it. Also don't know when anyone can give a proper review.

